PR #1, opened 2015-06-24
Convert to ES2015+ module

imurchie:
Working version of Remote Debugger module. Working on tests.


lib/handlers.js
```diff
@@ -0,0 +1,118 @@
...
+  }
+}
+
+async function handleDataMessage (plist) {
+  let dataKey = JSON.parse(plist.__argument.WIRMessageDataKey.toString('utf8'));
+  let msgId = dataKey.id;
+  let result = dataKey.result;
+  let error = dataKey.error || null;
+
+  if (error) throw new Error(error);
```
moizjv: I think we are doing away with one line if statements.

lib/callbacks.js
```diff
@@ -0,0 +1,45 @@
...
+  if (this.appIdKey === appIdKey) {
+    log.debug(`Page changed: ${JSON.stringify(pageDict)}`);
+    this.emit(RemoteDebugger.EVENT_PAGE_CHANGE, pageArrayFromDict(pageDict));
+  } else {
+    log.debug(`Received page change notice for app ${appIdKey} ` +
+              `but listening for ${this.appIdKey}. Ignoring.`);
+  }
+}
+
+async function onAppConnect (dict) {
```
jlipps: why is this an async function if it's (a) not returning a promise or (b) awaiting anything

lib/callbacks.js
```diff
@@ -0,0 +1,45 @@
...
+  let appIdKey = dict.WIRApplicationIdentifierKey;
+  log.debug(`Notified that a new application ${appIdKey} has connected`);
+
+  this.updateAppsWithDict(dict);
+}
+
+async function onAppDisconnect (dict) {
```
jlipps: same here

lib/handlers.js
```diff
@@ -0,0 +1,118 @@
...
+  if (fn) {
+    // most responses are only to be called once, then
+    // removed. But not the ones below, which handle
+    // page change and app connect/disconnect
+    if (specialCb !== '_rpc_forwardGetListing:' &&
+        specialCb !== '_rpc_applicationDisconnected:' &&
+        specialCb !== '_rpc_applicationConnected:') {
+      this.specialCbs[specialCb] = null;
+    }
+    fn(...args);
```
jlipps: yay spread

lib/handlers.js
jlipps: I'm a bit divided on this. It's similar to our old exception to the rule of no one-line if statements: short-circuiting and throwing an error.

Personally I'm ok with this as long as there's no `else`, but for the sake of consistency I'm also fine with using braces

lib/handlers.js
```diff
@@ -0,0 +1,118 @@
...
+  if (msgId !== null && typeof msgId !== 'undefined') {
+    msgId = msgId.toString();
+  }
+  if (dataKey.method === 'Profiler.resetProfiles') {
+    log.debug('Device is telling us to reset profiles. Should probably ' +
+              'do some kind of callback here');
+  } else if (dataKey.method === 'Page.frameNavigated') {
+    if (!this.willNavigateWithoutReload && !this.pageLoading) {
+      log.debug('Frame navigated, unloading page');
+      this.emit(RemoteDebugger._FRAME_NAVIGATED, true);
```
jlipps: does the underscore here mean this is an event we don't want a user to listen to, but one we use internally?

lib/handlers.js
```diff
@@ -0,0 +1,118 @@
...
+  } else if (this.dataCbs[msgId] === null) {
+    log.error(`Debugger returned data for message ${msgId} ` +
+              `but we already ran that callback! WTF??`);
+  } else {
+    if (!msgId && !result && !error) {
+      log.debug('Got a blank data response from debugger');
+    } else {
+      log.error(`Debugger returned data for message ${msgId} ` +
+                `but we were not waiting for that message! ` +
+                `result: ${JSON.stringify(result)} ` +
```
jlipps: do we want no punctuation between result and error? not even a little comma or semicolon? :-)

lib/helpers.js
```diff
@@ -0,0 +1,119 @@
...
+  };
+
+  return [id, entry];
+}
+
+function pageArrayFromDict (pageDict) {
+  let newPageArray = [];
+  for (let dict of _.values(pageDict)) {
+    // count only WIRTypeWeb pages and ignore all others (WIRTypeJavaScript etc)
+    if (_.isUndefined(dict.WIRTypeKey) || dict.WIRTypeKey === 'WIRTypeWeb') {
```
jlipps: I think we should either use `typeof ... 'undefined'` checks or lodash `isUndefined` checks in a project, but not both

lib/helpers.js
```diff
@@ -0,0 +1,119 @@
...
+      newPageArray.push({
+        id: dict.WIRPageIdentifierKey,
+        title: dict.WIRTitleKey,
+        url: dict.WIRURLKey,
+        isKey: !_.isUndefined(dict.WIRConnectionIdentifierKey),
```
jlipps: does lodash have `_.isDefined`?

lib/helpers.js
```diff
@@ -0,0 +1,119 @@
...
+    for (let frame of frames) {
+      script = wrapScriptForFrame(script, frame);
+    }
+  } else {
+    log.debug(`Executing '${atom}' atom in default context`);
+    script = `(${atomSrc})`;
+  }
+
+  // add the arguments, as strings
+  args = _.map(args, JSON.stringify);
```
jlipps: with ES6 can we not just use `args.map` instead of `_.map(args`?

lib/helpers.js
```diff
@@ -0,0 +1,119 @@
...
+  if (asyncCallBack ) {
```
jlipps: extra space here

lib/helpers.js
```diff
@@ -0,0 +1,119 @@
...
+
+function truncateForPrinting (str, length = 100) {
+  let truncated = str;
+  if (str.length > length) {
+    truncated = `${truncated.slice(0, length)}...`;
+  }
+  return truncated;
+}
+
+export { appInfoFromDict, pageArrayFromDict, getDebuggerAppKey, checkParams, wrapScriptForFrame, getScriptForAtom, truncateForPrinting };
```
jlipps: can we line up second lines below the first line, as in:

```js
export { foo, bar, baz, ...
         lol, ... };
```

lib/remote-debugger.js
```diff
@@ -0,0 +1,472 @@
...
+    });
+    this.socket.on('end', () => {
+      this.connected = false;
+    });
+    this.socket.on('data', this._receive.bind(this));
+
+    this.connected = true;
+
+    // connect the socket
+    let port = process.env.REMOTE_DEBUGGER_PORT || REMOTE_DEBUGGER_PORT;
```
jlipps: in addition to reading the env var, I believe we should also allow overriding the port in the constructor opts hash, so that we can have it sent in via appium later on. I don't think we should ever have gone the env var route with this, fwiw.

lib/remote-debugger.js
```diff
@@ -0,0 +1,472 @@
...
+    let appInfo = await this.setConnectionKey();
+    log.debug(`Received application data: ${JSON.stringify(appInfo)}`);
+    return appInfo;
+  }
+
+  async disconnect () {
+    log.debug('Disconnecting from remote debugger');
+    this.connected = false;
+    let asyncSocketDestroy = B.promisify(this.socket.destroy, this.socket);
+    await asyncSocketDestroy();
```
jlipps: can't decide if this extra assignment is useful for readability or not. is it better than just:

```js
await B.promisify(this.socket.destroy, this.socket)();
```

lib/remote-debugger.js
```diff
@@ -0,0 +1,472 @@
...
+        if (correctAppIdKey !== oldAppIdKey) {
+          log.debug(`We were notified we might have connected to the wrong app. ` +
+                    `Using id ${correctAppIdKey} instead of ${oldAppIdKey} ` +
+                    `and trying again`);
+          this.updateAppsWithDict(dict);
+          this.appIdKey = getDebuggerAppKey(this.bundleId, this.platformVersion, this.appDict);
+        }
+
+        // then try the whole thing again
+        this.selectApp(maxTries, tries + 1);
```
jlipps: i'm confused why we don't need to `await` or `return` this

lib/remote-debugger.js
```diff
@@ -0,0 +1,472 @@
...
+    if (_.isEmpty(pageDict)) {
+      let msg;
+      if (this.appIdKey === appIdKey) {
+        msg = 'Empty page dictionary received. Trying again.';
+      } else {
+        // ignore apps we aren't listening for
+        msg = `Received empty page dictionary for app ${appIdKey} ` +
+              `but listening for ${this.appIdKey}. Ignoring.`;
+      }
+      return requiresRetryCb(this.appIdKey, this.appIdKey, msg);
```
jlipps: is there not a way to use `asyncbox.retry` so you don't need to handle the try count logic?

lib/remote-debugger.js
```diff
@@ -0,0 +1,472 @@
...
+      if (ready || (this.pageLoadMs > 0 && (start + this.pageLoadMs) < Date.now())) {
+        log.debug('Page is ready');
+        clearTimeout(this.loadingTimeout);
+        this.pageLoading = false;
+        d.resolve();
+      } else {
+        log.debug('Page was not ready, retrying');
+        this.loadingTimeout = setTimeout(verify, timeoutMs);
+      }
+    }.bind(this);
```
jlipps: shouldn't need `.bind(this)` with `=>`

lib/remote-debugger.js
```diff
@@ -0,0 +1,472 @@
...
+  }
+
+  async selectApp (maxTries = SELECT_APP_RETRIES, tries = 1) {
+    // try to get the app id from our connected apps
+    this.appIdKey = getDebuggerAppKey(this.bundleId, this.platformVersion, this.appDict);
+
+    // local callback, temporarily added as callback to
+    // `_rpc_applicationConnected:` remote debugger response
+    // to handle the initial connection
+    let requiresRetryCb = async (dict, msg) => {
```
sebv: async

lib/remote-debugger.js
```diff
@@ -0,0 +1,472 @@
...
+      if (errors) throw new Error(errors);
+    }
+
+    log.debug(`Navigating to new URL: ${url}`);
+    let navToUrlCmd = RemoteMessages.setUrl(url, this.appIdKey, this.connId,
+                          this.senderId, this.pageIdKey, this.debuggerType);
+    await this._send(navToUrlCmd);
+
+    // Promise to be resolved after we wait for the frame
+    let d = B.pending();
```
jlipps: would rather see this all as:

```js
if (!this.useNewSafari) {
  await B.delay(1000);
}
await this.waitForFrameNavigated();
```

no more requirement of using promise/resolve/etc.

moral of the story: we should be very suspicious when we are creating a promise and rejecting/resolving that there's not a pure async/await way of doing it.

lib/remote-debugger.js
sebv: await?

lib/remote-debugger.js
```diff
@@ -0,0 +1,472 @@
...
+
+    log.debug('Waiting for frame navigated...');
+    var startMs = Date.now();
+
+    // wait for the FRAME_NAVIGATED event,
+    // which will come from a timeout or from a callback
+    // when remote debugger returns a `Page.frameNavigated`
+    // response
+    this.on(RemoteDebugger._FRAME_NAVIGATED, function () {
```
jlipps: here's a case where we do need it, since we have to listen for an external event using a callback

lib/remote-debugger.js
```diff
@@ -0,0 +1,472 @@
...
+      // only need to do anything if the promise is still pending
+      if (d.promise.isPending()) {
+        log.debug(`Frame navigated in ${((Date.now() - startMs)/1000)} sec.`);
+        clearTimeout(this.navigatingTimeout);
```
jlipps: I'm not convinced that `this` is what we think it is here, since we didn't use `=>` above

lib/remote-debugger.js
```diff
@@ -0,0 +1,472 @@
...
+  async stopTimeline () {
+    log.debug('Stopping to record the timeline');
+    let sendJSCommandCmd = RemoteMessages.stopTimeline(this.appIdKey,
+                              this.connId, this.senderId, this.pageIdKey,
+                              this.debuggerType);
+    return await this._send(sendJSCommandCmd);
+  }
+
+  async timelineEventRecorded (result) {
+    this.timelineEventHandler(result);
```
jlipps: what's the purpose of a one-line function? why not just call the inner line directly?

jlipps: also unclear why this would be an async function

lib/remote-debugger.js
```diff
@@ -0,0 +1,472 @@
...
+RemoteDebugger.prototype.setHandlers = setHandlers;
+
+// generic callbacks
+RemoteDebugger.prototype.pageChangeCb = pageChangeCb;
+RemoteDebugger.prototype.onAppConnect = onAppConnect;
+RemoteDebugger.prototype.onAppDisconnect = onAppDisconnect;
+
+
+// factory method for creating a RemoteDebugger instance with the correct values
+function init (bundleId, platformVersion, useNewSafari, pageLoadMs) {
```
jlipps: little unclear to me what the utility of this `init` method is.

lib/remote-debugger.js
```diff
@@ -0,0 +1,472 @@
...
+    this.connected = false;
+    this.connId = uuid.v4();
+    this.senderId = uuid.v4();
+    this.curMsgId = 0;
+    this.dataCbs = {};
+    this.received = new Buffer(0);
+    this.readPos = 0;
+
+    // set up the special callbacks for handling rd events
+    this.specialCbs = {
```
sebv: Should we rename all those cbs?  Maybe rpcHandlers would be better.

lib/socket.js
```diff
@@ -0,0 +1,164 @@
...
+      return;
+    }
+
+    // bplistParse.parseBuffer returns an array
+    if (plist.length === 1) {
+      plist = plist[0];
+    }
+
+    var plistCopy = plist;
+    if (typeof plistCopy.WIRMessageDataKey !== "undefined") {
```
jlipps: some duplication of logic here, what about:

```js
for (let key of ['WIRMessageDataKey', 'WIRDestinationKey', 'WIRSocketDataKey']) {
  if (typeof plistCopy[key] !== "undefined") {
    plistCopy[key] = plistCopy[key].toString("utf8");
  }
}
```

will save a few lines and make easier to maintain

lib/webkit-remote-debugger.js
```diff
@@ -0,0 +1,193 @@
...
+    log.debug('Getting WebKitRemoteDebugger pageArray');
+    let pageElementJSON = await this.getJSONFromUrl(this.host, this.port, '/json');
+    log.debug(`Page element JSON: ${pageElementJSON}`);
+    // Add elements to an array
+    let newPageArray = [];
+    for (let pageObject of pageElementJSON) {
+      let urlArray = pageObject.webSocketDebuggerUrl.split('/').reverse();
+      let id = urlArray[0];
+      newPageArray.push({
+        id: id,
```
jlipps: `id,`

lib/webkit-remote-debugger.js
```diff
@@ -0,0 +1,193 @@
...
+        title: pageObject.title,
+        url: pageObject.url,
+        isKey: id,
```
jlipps: maybe `!!id` to make it clear what we are doing here?

lib/webkit-remote-debugger.js
```diff
@@ -0,0 +1,193 @@
...
+    }
+    return newPageArray;
+  }
+
+  async getJSONFromUrl (host, port, path) {
+    // only resolve once the GET request has been responded to
+    let d = B.pending();
+
+    // retrieve data from the WebKit instance
+    http.get({ host: host, port: port, path: path }, function (response) {
```
jlipps: no need for spaces in braces here, also:

```js
{host, port, path}
``

Comment (jlipps) [inline lib/webkit-remote-debugger.js]: why not use `request-promise` instead of reimplementing w/ http? then we don't need to make our own promise wrapper

**Comment by jlipps** [comment]
wow, this was a big one! really great translation, and it's already much, much easier to read and understand.

i made a number of comments. there might be more places we can simplify even further w/ various async/await tricks, but i pointed out the ones that i saw.

File: lib/remote-debugger.js [hunk 28]
```diff
@@ -0,0 +1,472 @@
...
+  }
+}
+
+// event emitted publically
+RemoteDebugger.EVENT_PAGE_CHANGE = 'remote_debugger_page_change';
+
+// internal event
+RemoteDebugger._FRAME_NAVIGATED = 'remote_debugger_frame_navigated';
+
+// socket stuff
```
sebv: I think we need to break the pattern below, either create mixin, or better, use straightforward composition. You see this in a lot of places in Appium code.

lib/remote-debugger.js
sebv: There is a timeout function in Bluebird to do the same as setTimeout.

lib/remote-debugger.js
```diff
@@ -0,0 +1,472 @@
...
+    this.bundleId = bundleId;
+    this.platformVersion = platformVersion;
+    this.debuggerType = debuggerType || DEBUGGER_TYPES.webinspector;
+    if (this.debuggerType === DEBUGGER_TYPES.webinspector) {
+      this.useNewSafari = useNewSafari || false;
+      this.pageLoadMs = pageLoadMs;
+      log.debug(`useNewSafari --> ${this.useNewSafari}`);
+    }
+
+    // app handling configuration
```
sebv: It would be good to organize the data better (see unix tao) , and loose the habit of having classes with too many fields. (also general remark about old appium code.)

lib/handlers.js
imurchie: That's the idea, yes.

lib/helpers.js
imurchie: I would agree. And I would be in the `_.isUndefined` camp.

lib/helpers.js
imurchie: Not that I can see.

lib/helpers.js
imurchie: Yes. That was something I had intended to do, but had been waiting for the list to settle down. Alas.

lib/remote-debugger.js
imurchie: Both good questions. ;)

lib/remote-debugger-rpc-client.js
```diff
@@ -0,0 +1,246 @@
...
+    this.host = host;
+    this.port = port;
+    this.connected = false;
+    this.connId = uuid.v4();
+    this.senderId = uuid.v4();
+    this.curMsgId = 0;
+    this.received = new Buffer(0);
+    this.readPos = 0;
+
+    // this.messageHandler = new RpcMessageHandler(specialRpcHandlers);
```
jlipps: if you're not using this line, let's axe it

lib/remote-debugger-rpc-client.js
```diff
@@ -0,0 +1,246 @@
...
+      this.connected = true;
+
+      d.resolve();
+    });
+    this.socket.on('error', (exception) => {
+      if (this.connected) {
+        log.error(`Socket error: ${exception.message}`);
+        this.connected = false;
+      }
+      if (d.promise.isPending() && exception.code === 'ECONNREFUSED') {
```
jlipps: what happens if we get any other kind of exception?

also, do we need the `isPending` check? worst case scenario we reject after it's been resolved, which does (afaik) nothing

lib/remote-debugger-rpc-client.js
```diff
@@ -0,0 +1,246 @@
...
+  allowNavigationWithoutReload (allow = true) {
+    this.messageHandler.allowNavigationWithoutReload(allow);
+  }
+
+  async send (command, opts = {}) {
+    // promise to be resolved whenever remote debugger
+    // replies to our request
+    let d = B.pending();
+
+    let data = getRemoteCommand(command, _.defaults({connId: this.connId, senderId: this.senderId}, opts));
```
jlipps: line looks long (we could construct the object literal on the line above)

lib/remote-debugger-rpc-client.js
```diff
@@ -0,0 +1,246 @@
...
+    } catch (e) {
+      log.error(`Could not create binary plist from data: ${e}`);
+      return;
+    }
+
+    let onSocketError = (exception) => {
+      if (this.connected) {
+        log.error(`Socket error: ${exception.message}`);
+      }
+      if (d.promise.isPending()) {
```
jlipps: same q here about `isPending`

lib/remote-debugger.js
```diff
@@ -0,0 +1,432 @@
...
+        log.debug('Page is ready');
+        clearTimeout(this.loadingTimeout);
+        this.pageLoading = false;
+        d.resolve();
+      } else {
+        log.debug('Page was not ready, retrying');
+        this.loadingTimeout = setTimeout(verify, timeoutMs);
+      }
+    };
+    this.loadingTimeout = setTimeout(verify, timeoutMs);
```
jlipps: still not entirely sure why we are using `setTimeout` instead of `await B.delay(timeoutMs);` then continuing with the logic.

lib/remote-debugger.js
```diff
@@ -0,0 +1,432 @@
...
+
+    return d.promise;
+  }
+
+  async cancelPageLoad () {
```
jlipps: oh is it because we might need to do this?

lib/remote-debugger.js
```diff
@@ -0,0 +1,432 @@
...
+    log.debug(`Navigating to new URL: ${url}`);
+    await this.rpcClient.send('setUrl', {
+      url: url,
+      appIdKey: this.appIdKey,
+      pageIdKey: this.pageIdKey,
+      debuggerType: this.debuggerType
+    });
+
+    if (!this.useNewSafari) {
+      await B.delay(1000);
```
jlipps: this static wait probably deserves a comment

lib/remote-debugger.js
```diff
@@ -0,0 +1,432 @@
...
+      debuggerType: this.debuggerType
+    });
+  }
+
+  async callFunction (objId, fn, args) {
+    let errors = checkParams({appIdKey: this.appIdKey, pageIdKey: this.pageIdKey});
+    if (errors) throw new Error(errors);
+
+    log.debug('Calling javascript function');
+    return await this.rpcClient.send('callJSFunction', {
```
jlipps: i think it's ok to just `return` instead of `return await`? we probably need to settle on this as a team

lib/webkit-rpc-client.js
```diff
@@ -0,0 +1,150 @@
...
+    log.debug(`Sending WebKit data: ${truncateForPrinting(JSON.stringify(data), 200)}`);
+
+    this.curMsgId++;
+    data.id = this.curMsgId;
+
+    // only resolve the send command when WebKit returns a response
+    let d = B.pending();
+
+    // store the call back and the data sent
+    this.dataCbs[this.curMsgId.toString()] = d.resolve.bind(d); // dataCb;
```
jlipps: don't understand the `dataCb` comment

jlipps:
:+1: overall this new more stratified structure is clearer and easier to understand. made a few more comments

lib/remote-debugger-rpc-client.js
sebv: You can just put it in a `else` (i.e if not connected.).

lib/remote-debugger-rpc-client.js
```diff
@@ -0,0 +1,246 @@
...
+      this.socket = null;
+    });
+    this.socket.on('end', () => {
+      this.connected = false;
+    });
+    this.socket.on('data', this.receive.bind(this));
+
+    // connect the socket
+    // only resolve this function when we are actually connected
+    let d = B.pending();
```
sebv: Should we use the documented way to create promise. That works (reading the source), but cannot find it the api doc.

lib/remote-debugger-rpc-client.js
```diff
@@ -0,0 +1,246 @@
...
+        d.reject(exception);
+      }
+    });
+    return d.promise;
+  }
+
+  async disconnect () {
+    log.debug('Disconnecting from remote debugger');
+    this.connected = false;
+    await B.promisify(this.socket.destroy, this.socket)();
```
sebv: Looks like destroy has no callback: https://nodejs.org/api/net.html#net_socket_destroy

lib/remote-debugger.js
sebv: There B.timeout, it is cancellable.

lib/remote-debugger.js
sebv: bluebird can handle this.

lib/webkit-rpc-client.js
sebv: another variable to rename.

lib/remote-debugger-rpc-client.js
imurchie: Do you mean wrapping it all in a `new Promise()` call?

sebv: yes it is deprecated: see here: https://github.com/petkaantonov/bluebird/blob/master/API.md#deferred-migration

lib/remote-debugger.js
imurchie: I guess I'm just not seeing what the benefit of doing it with Bluebird would be.

sebv: You don't need the resolve.

sebv: And in your case verify returns a promise, and you don't hook it anywhere, so exception handling is not great.

sebv: @imurchie see gist. 
https://gist.github.com/sebv/20c8d08d3a71a98c21c4

imurchie:
Addressed the comments and I think it is in a good state now.

lib/helpers.js
```diff
@@ -0,0 +1,136 @@
...
+
+function truncateForPrinting (str, length = 100) {
+  let truncated = str;
+  if (str.length > length) {
+    truncated = `${truncated.slice(0, length)}...`;
+  }
+  return truncated;
+}
+
+function cancellableDelay (ms) {
```
jlipps: this is a cool pattern. maybe should become part of some util package?

PR #4, opened 2015-08-27
Return values, not remote debugger objects

imurchie:
Don't allow the remote debugger objects bleed out into client code.


lib/remote-debugger.js
```diff
@@ -385,6 +381,33 @@ class RemoteDebugger extends events.EventEmitter {
...
+        // we might get a serialized object, but we might not
+        // if we get here, it is just a value
+      }
+    } else if (!_.isObject(res)) {
+      throw new Error(`Result has unexpected type: (${typeof res}).`);
+    }
+
+    if (res.status && res.status !== 0) {
+      // we got some form of error.
+      console.log(res);
```
jlipps: probably want to log.debug or something instead of console.log?

imurchie: There's always one! It must be a law or something.

PR #5, opened 2015-09-01
Fix handling of waiting for dom

imurchie:
We need to wait for the dom at one point. Also fix error


lib/remote-debugger.js
```diff
@@ -401,7 +402,8 @@ class RemoteDebugger extends events.EventEmitter {
 
     if (res.status && res.status !== 0) {
       // we got some form of error.
-      throw new errors.JavaScriptError(`${res.value} (status: ${res.status})`);
+      let message = res.value.message ? res.value.message : res.value;
```
jlipps: what about

```js
let message = res.value.message || res.value;
```

PR #10, opened 2015-10-26
Do not disconnect when no apps are present

imurchie:
Rather than erroring when there are no apps present, just keep going. Otherwise reconnecting too soon causes the remote debugger on the device to freak out and stop responding. The client will disconnect when the client feels like it.

@scottdixon This, in addition to some changes once this lands, fixes the getting context multiple times issue.

### DIFF
--- a/lib/remote-debugger.js
+++ b/lib/remote-debugger.js
@@ -30,7 +30,7 @@ class RemoteDebugger extends events.EventEmitter {
    *   - host - the remote debugger's host address
    *   - port - the remote debugger port through which to communicate
    */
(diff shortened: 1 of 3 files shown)


lib/remote-debugger.js
```diff
@@ -86,6 +86,10 @@ class RemoteDebugger extends events.EventEmitter {
     this.emit(RemoteDebugger.EVENT_DISCONNECT, true);
   }
 
+  isConnected () {
```
jlipps: could also have implemented these as `get`ters.

PR #11, opened 2015-11-20
Keep track of updated remote debugger app

christian-bromann:
We have had some issues with iOS 9+ and webviews connecting then hanging. Fix this by remembering the actually connected app.

Original PR: appium/appium#5800


christian-bromann:
build started failing after applying the fix.
Reason: `AssertionError: expected 'PID:42' to equal 'PID:44'`

imurchie:
That's to be expected, since we're now changing the PID on a redirect. This line needs to be changed to reflect that:

https://github.com/appium/appium-remote-debugger/blob/master/test/unit/remote-debugger-specs.js#L117

PR #13, opened 2015-12-14
Reset page dictionary if url is wrong

imurchie:
Rather than letting it fall through and act like a page was actually selected if all the retries are used up. This way we throw the proper error.


jlipps:
:+1:

PR #14, opened 2015-12-16
Fix handling of application disconnects

imurchie:
Rather than always finding a new app id, only do it if we've actually lost the one we are connected to.


jlipps:
:+1:

PR #22, opened 2016-05-12
Clean up special message handling

imurchie:
Wrap and upwrap the special message handler so we don't lose it in the course of operating. Also clean up handling of error listener.


jlipps:
👍  pretty gnarly!

PR #23, opened 2016-05-19
Better mapping of pages, to allow multiple wkwebviews on sims

imurchie:
Traditional webviews return an app with multiple pages associated with it. We would find the one we wanted. In wkwebviews each page is mapped to its own app. So rather than looking at a single app, we need to look at all of them.

This changes the webview ids from simple integers (which were app ids) to composites of the app and the page. For UIWebviews the webviews will be, for app `PID:1234` with four pages, something like `1234.1`, `1234.2`, `1234.3`, and `1234.4`. For WkWebviews we will get ids like `1234.0`, `1235.0`, `1236.0`, and `1237.0`.

Since there is no _a priori_ way to tell what the id would be for a particular webview in the previous scheme, this should not be a breaking change.

See discussion https://github.com/appium/appium/issues/5839

### DIFF
--- a/lib/remote-debugger.js
+++ b/lib/remote-debugger.js
@@ -6,7 +6,7 @@ import { errors } from 'mobile-json-wire-protocol';
 import RemoteDebuggerRpcClient from './remote-debugger-rpc-client';
 import messageHandlers from './message-handlers';
 import { appInfoFromDict, pageArrayFromDict, getDebuggerAppKey, getPossibleDebuggerAppKeys, checkParams,
-         getScriptForAtom, simpleStringify } from './helpers';
+         getScriptForAtom, simpleStringify, deferredPromise } from './helpers';
 import { util } from 'appium-support';
 import _ from 'lodash';
 import Promise from 'bluebird';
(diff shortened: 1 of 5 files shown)


lib/remote-debugger.js
```diff
@@ -139,6 +140,19 @@ class RemoteDebugger extends events.EventEmitter {
     let [id, entry] = appInfoFromDict(dict);
     this.appDict[id] = entry;
 
+    // add a promise to get the page dictionary
+    if (_.isUndefined(entry.pageDict)) {
+      // we want to be able to resolve the promise from afar, so save the resolve function
+      let res;
```
jlipps: this section seems a little tortured to me. in fact I don't even know what's going on here.

jlipps:
hooray for more complexity in the remote debugger!

lib/remote-debugger.js
imurchie: :(

Essentially, we get notified of applications, and then at a later time we get notifications of the pages associated with each app. But we don't want to return from seeing what apps are avaialble until after the pages are in. So we have a promise for each app's pages, which is resolved when the page comes in. Later we can wait for them all to be fulfilled.

The tortured part is because, as far as I can see, there is no way to resolve a promise from without. So we're saving the resolve function on the promise itself. 

Bleh.

jlipps: ok. yes i think you're right that we need to use the deferred pattern here. does this simplification work at all?

```js
let p = new Promise((resolve) => {
  p.resolve = resolve;
});
entry.pageDict = p;
```

jlipps: or not since the Promise constructor is called before the assignment?

PR #24, opened 2016-07-01
Check for the right app better

imurchie:
Rather than simply looking through the current record, look through all the records we have already to see if the page has been found in the background. This should (a) speed things up and (b) make things more reliable, as we don't need to interact with the remote debugger so much.

### DIFF
--- a/lib/remote-debugger.js
+++ b/lib/remote-debugger.js
@@ -149,7 +149,9 @@ class RemoteDebugger extends events.EventEmitter {
     this.logApplicationDictionary(this.appDict);
 
     // try to get the app id from our connected apps
-    this.appIdKey = getDebuggerAppKey(this.bundleId, this.platformVersion, this.appDict);
+    if (this.appIdKey) {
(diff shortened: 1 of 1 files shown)


lib/remote-debugger.js
```diff
@@ -182,31 +182,32 @@ class RemoteDebugger extends events.EventEmitter {
...
+
+          // if we are looking for a particular url, make sure we have the right page
           if (currentUrl) {
-            let found;
-            _.each(pageDict, (record) => {
-              if (record.WIRURLKey === currentUrl) {
-                // this is the page we want
-                found = true;
+            let found = false;
+            for (let appDict of _.values(this.appDict)) {
```
jlipps: I think this might be more elegantly written as:

```js
outer: for (let appDict of _.values(this.appDict)) {
  for (let dict of (appDict.pageDict || [])) {
    if (dict.url === currentUrl) {
      found = true;
      break outer;
    }
  }
}
```

jlipps: (http://stackoverflow.com/questions/1564818/how-to-break-nested-loops-in-javascript)

PR #28, opened 2016-08-08
Add support for iOS 10

imurchie:
Mostly this is cosmetic. The crux of the matter is adding `WIRAutomaticallyPause: false` to the arguments sent on the `_rpc_forwardSocketSetup:` call (setting the sending id before choosing the page). Without this the connection to the app succeeds, but almost immediately the remote debugger disconnects the app and we are left with no way to automate it.

This appears to be fully backward compatible.

### DIFF
--- a/lib/remote-debugger.js
+++ b/lib/remote-debugger.js
@@ -11,6 +11,7 @@ import { util } from 'appium-support';
 import _ from 'lodash';
 import Promise from 'bluebird';
 
+
 const DEBUGGER_TYPES = {
   webkit: 1,
   webinspector: 2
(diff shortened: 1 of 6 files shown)


lib/remote-debugger.js
```diff
@@ -232,9 +238,11 @@ class RemoteDebugger extends events.EventEmitter {
            this.onPageChange.bind(this));
 
     // wait for all the promises are back, or 30s passes
-    let pagePromises = _.map(this.appDict, (app) => app.pageDict && app.pageDict.promise);
-    log.debug(`Waiting for ${pagePromises.length} pages to be fulfilled`);
-    await Promise.any([Promise.delay(30000), Promise.all(pagePromises)]);
+    let pagePromises = _.map(_.filter(this.appDict, (app) => !!app.pageDict && !!app.pageDict.promise), (app) => app.pageDict.promise);
```
jlipps: aren't `map` and `filter` proper methods now? might look nicer that way and chained

jlipps:
one comment then LGTM

PR #29, opened 2016-08-11
Make sure we keep track of the page we have chosen

imurchie:
When we check the apps we are breaking out of the selection when we have the right app, but not keeping track of what that app is. This means we end up getting _an_ app but not necessarily the _right_ app.

So keep track of the id and the page dictionary where the page is.


jlipps:
ahhhh

imurchie:
This is the worst part of the multiple-package strategy. I have a problem that only manifests in Travis, in another package. So I can't debug it without fixing a package, publishing it, then fixing in the other package and pushing it. Ugh.

PR #30, opened 2016-08-26
Make sure we cleanup handlers when disconnecting

imurchie:
Otherwise we can get events after we take the session as over.


jlipps:
👍  do we need to publish this still?

PR #33, opened 2016-09-07
Change direction of condition

imurchie:
Condition was backwards, so sometimes we would end up, inexplicably, in the wrong app and timeout.


jlipps:
I _thought_ something was kinda wonky with that. This code is still so opaque :-(

imurchie:
Yes, it leaves much to be desired. Hopefully mobile safari will go [the way of desktop safari](https://webkit.org/blog/6900/webdriver-support-in-safari-10/)

PR #34, opened 2016-09-08
fix the truncate usage

bigfei:
According to the [doc](https://lodash.com/docs/4.15.0#truncate), length should be in Object notation.

### DIFF
--- a/lib/remote-debugger-rpc-client.js
+++ b/lib/remote-debugger-rpc-client.js
@@ -171,7 +171,7 @@ export default class RemoteDebuggerRpcClient {
         // temporarily wrap with promise handling
         let specialMessageHandler = this.getSpecialMessageHandler(data.__selector);
         this.setSpecialMessageHandler(data.__selector, reject, function (...args) {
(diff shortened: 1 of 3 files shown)


bigfei:
@imurchie 
It seems the jshint cannot lint the "{}" inside the string template.
eg.
```javascript
var s = `${{}.toString()}`
console.log(s)
```
It outputs a error like ` Expected '}' to match '{' from line 1 and instead saw '}.toString()}'.`
So i suggest trying to move into the new eslint-config-appium as the linter. 

I'd very pleased to do the eslint migration  if it is possible/within the schedule.

imurchie:
Two options: 

(1) add `// jshint ignore:line` to the line.

(2) add the length as a constant in `lib/logger.js` which is then exported

```js
import { getLogger } from 'appium-logger';

const log = getLogger('RemoteDebugger');
const LENGTH = {length: 50};

export { log, LENGTH };
export default log;
```

Then

```js
import { log, LENGTH };

// ...

log.debug(`Received response from socket send: '${_.truncate(JSON.stringify(args), LENGTH)}'`);
```

PR #35, opened 2016-09-08
putting null and undefined check for error message

yaphatak:


lib/remote-debugger-message-handler.js
```diff
@@ -83,9 +83,11 @@ export default class RpcMessageHandler {
     let error = dataKey.error || null;
 
     // we can get an error, or we can get a response that is an error
-    if (result && result.wasThrown) {
+    if (result && result.wasThrown && result.result) {
```
imurchie: I think this check is in the wrong place. We want to error out even if there is no `result.result`. I know that wasn't how it was before, because of the checks, but I think we ought to throw a generic error if `result.wasThrown` is true but there is no message.

yaphatak: For me error was thrown on `message` being undefined. I added `result.result` as just extra check. Let me remove that or check different way

yaphatak: But What i saw, it did return pagesource but because of `message` being undefined it blew up there.

lib/remote-debugger-message-handler.js
```diff
@@ -84,8 +84,13 @@ export default class RpcMessageHandler {
 
     // we can get an error, or we can get a response that is an error
     if (result && result.wasThrown) {
-      let message = result.result.value || result.result.description;
-      error = new Error(message);
+      let message = `Error occurred in handling data message`;
+      if (result.result) {
+         message = result.result.value || result.result.description;
```
imurchie: I think we ought to default even if `result.result.value` _and_ `result.result.description` are blank. Something like

```js
let message = (result.result && (result.result.value || result.result.description)) ? 
                           (result.result.value || result.result.description) :
                           'Error occurred in handling data message';
```

That way when `result.wasThrown` is true, we throw an error. no matter what.

yaphatak: this `plist` is response from apple webinspector? and `wasThrown` comes as part of it which we trust?

imurchie: Trust? Not really. It is one of a few ways the inspector returns an error.

yaphatak: okay. I updated PR with your suggestion.

imurchie:
This looks good to me!

imurchie:
Published in `appium-remote-debugger@3.1.6`

PR #36, opened 2016-09-09
don't connect to app if url is `about:blank` or empty url

yaphatak:
What I observed is remote debugger hangs sometimes if connected to webview with empty url or about:blank. May be because at that point it is still trying to load page. Here I totally neglect empty url and have option of ignoring `about:blank` so that it doesn't hang anymore.

### DIFF
--- a/lib/remote-debugger.js
+++ b/lib/remote-debugger.js
@@ -189,7 +189,7 @@ class RemoteDebugger extends events.EventEmitter {
     }
   }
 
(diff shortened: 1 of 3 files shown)


imurchie:
I'm ok with this. I've not seen about:blank causing problems, but if it is, it is!

imurchie:
Published in `appium-remote-debugger@3.1.5`

PR #38, opened 2016-09-14
filter out pages with empty urls

yaphatak:


lib/remote-debugger.js
```diff
@@ -216,43 +219,46 @@ class RemoteDebugger extends events.EventEmitter {
           this.appDict[appIdKey].pageDict = pageArrayFromDict(pageDict);
 
           // if we are looking for a particular url, make sure we have the right page. Ignore empty or undefined urls. Ignore about:blank if requested.
-          let found = false;
           dictLoop: for (let appDict of _.values(this.appDict)) {
-            if (found) break;
+            if (foundAtLeastAnApp){
+              break;
```
imurchie: This can break from `appLoop` and then you don't need the stuff in line 202-204, either. See how it works in line 249 below.

lib/remote-debugger.js
```diff
@@ -216,43 +219,46 @@ class RemoteDebugger extends events.EventEmitter {
...
+            }
             for (let dict of (appDict.pageDict || [])) {
               if (dict.url && (!ignoreAboutBlankUrl || dict.url !== 'about:blank') && (!currentUrl || dict.url === currentUrl)) {
                 // save where we found the right page
-                appIdKey = appDict.id;
+                foundAtLeastAnApp = true;
                 pageDict = dict;
-                found = true;
+                appIdKey = appDict.id;
+                if (this.appIdKey !== appDict.id) {
```
imurchie: Might as well use `appIdKey` instead of `appDict.id` now that it is set.

lib/remote-debugger.js
```diff
@@ -266,21 +272,6 @@ class RemoteDebugger extends events.EventEmitter {
...
-      if (!_.isArray(info.pageDict)) continue;
-      let id = app.replace('PID:', '');
-      for (let page of info.pageDict) {
-        let pageDict = _.clone(page);
-        pageDict.id = `${id}.${pageDict.id}`;
-        fullPageArray.push(pageDict);
-      }
-    }
-
     return fullPageArray;
```
imurchie: If this is relying on pushing the page array during the loop above (while receiving apps), this will _not_ work. We don't have all the information about that apps until **after** the promises are fulfilled for each app.

yaphatak: So does that mean, I still have to go through this 
`for (let [app, info] of _.toPairs(this.appDict)) {		
      if (!_.isArray(info.pageDict)) continue;		
      let id = app.replace('PID:', '');		
      for (let page of info.pageDict) {		
        let pageDict = _.clone(page);		
        pageDict.id = `${id}.${pageDict.id}`;		
        fullPageArray.push(pageDict);		
     }		
    }`

And then while going through this loop return pages only that machines with pushed app-pages during loop

imurchie: I'm not sure what you mean by the last comment.

imurchie: The problem is, after the main loop (`appLoop`), only one app has been put into `fullPageArray`, generally speaking. Any other apps have not yet been added. This is fine for WebViews, but WKWebViews will only ever, then, have one page reported (because in WKWebViews each page is represented by a different "app" in the debugger).

yaphatak: If you see i break out of `appLook` only if `selectOnlyFirstApp` && `foundAtLeastAnApp`.  `selectOnlyFirstApp` is `false` by default.

imurchie: Also if you get into the next iteration of `appLoop` and have found an app already (lines 202-204).

yaphatak: So, does that mean we can only get `WKWebViews` if we try for `maxTries` times. Sorry for long conversation.

jlipps:
@imurchie where are we at with this one?

yaphatak:
I think I have simple solution for this. I will send in different PR

PR #39, opened 2016-10-03
Remove requirement of having a non-blank url

imurchie:
Otherwise we cannot get webviews that do not have any content yet. Appium's webview tests fail with this check.


jlipps:
👍

Issue #42, opened 2016-10-19
Appium hangs on unresponsive webview


ssynix:
Our app uses WKWebviews and has built a number of custom functionality surrounding URL navigation. For example, if the app has no internet access, the webview navigates to a native page which displays the proper error message. Navigating to certain URLs within the webview also triggers the creation of a new webview and disabling the original webview.

Appium naturally does not expect this behavior, so I've resorted to handling URL navigation by executing javascript, as I don't believe it is logical for Appium to handle our convoluted webview navigation logic. The issue that I've run into is that if the app ends up in one of these unresponsive webviews, or if the URL navigation is interrupted by a native page, Appium hangs since IWDP never returns a response.

The simplest solution from my end is to [add a timeout](https://github.com/appium/appium-remote-debugger/blob/master/lib/webkit-rpc-client.js#L102) to the promise that's expecting a response from the webview, which results in an exception that I can catch and handle appropriately from my end.

Let me know if this is the right solution, and whether it has any potential side effects on the rest of the remote-debugger that I should be aware of. I'm definitely open to submitting the PR.

imurchie:
I think that sounds like a good idea, in general. If the promise is correctly, and the data handlers are handled correctly (probably best to move the `then` handler into a `finally` [here](https://github.com/appium/appium-remote-debugger/blob/master/lib/webkit-rpc-client.js#L94)), it should not have side effects.

rkavalap:
@ssynix  Closing this issue, since your PR for this is merged. If you believe something needs to be enhanced, can you open a new issue in main appium repo.

PR #43, opened 2016-10-20
Ensuring Appium times out gracefully if webkit does not return a response

ssynix:
Addressing #42.

I went with a simple 5 sec timeout. Let me know if it needs to be configured through a capability.

### DIFF
--- a/lib/remote-debugger.js
+++ b/lib/remote-debugger.js
@@ -19,6 +19,9 @@ const DEBUGGER_TYPES = {
 const SELECT_APP_RETRIES = 20;
 const REMOTE_DEBUGGER_PORT = 27753;
 
+/* How many milliseconds to wait for webkit to return a response before timing out */
+const RPC_RESPONSE_TIMEOUT_MS = 5000;
+
 
 class RemoteDebugger extends events.EventEmitter {
   /*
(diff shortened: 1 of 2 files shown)


lib/remote-debugger.js
```diff
@@ -19,6 +19,9 @@ const DEBUGGER_TYPES = {
 const SELECT_APP_RETRIES = 20;
 const REMOTE_DEBUGGER_PORT = 27753;
 
+/* How many milliseconds to wait for webkit to return a response before timing out */
+const RPC_RESPONSE_TIMEOUT_MS = 5000;
```
imurchie: If it is only used for Webkit, why not declare it there?

ssynix: Just so the constants are defined in the same place. Do you prefer I move it to the other file?

imurchie: No, I guess that makes some sense. :)

imurchie:
Sorry, I lost track of this. Thanks for the work in getting this in!

imurchie:
Published in `appium-remote-debugger@3.1.9`

ssynix:
Thanks! Will this make it to `1.6.1`?

imurchie:
It will. It should be in the beta.

PR #48, opened 2016-11-25
Do not select an app if its pageDict is invalid

cschabetsberger-micro-focus:
This adresses issue #45: It seems that Safari on iOS >= 10 simulator returns appIds for https://www.google.com, https://www.yahoo.com, ... (depends on which search engines are configured) even though only 1 tab is open in Safari for my own local page. This might have been also the case in earlier iOS versions but the order or returned appIds is different on every call. If Appium connects to such an appId, the connection may break down or may not even respond to a call. To detect invalid appIds, it seems to be sufficient to check if the pageId >= 5 of a page dict. Only invalid tabs seems to have such an id >= 5 which was the only indicator of an invalid/hidden appId/tab I could find.

### DIFF
--- a/lib/remote-debugger-rpc-client.js
+++ b/lib/remote-debugger-rpc-client.js
@@ -126,6 +126,11 @@ export default class RemoteDebuggerRpcClient {
         appIdKey
       });
 
+      // Reject the page dict if it is not valid
+      if (!this.isValidPageDict(pageDict)) {
+        pageDict = null;
+      }
+
       // sometimes the connect logic happens, but with an empty dictionary
       // which leads to the remote debugger getting disconnected, and into a loop
       if (_.isEmpty(pageDict)) {
@@ -139,6 +144,19 @@ export default class RemoteDebuggerRpcClient {
       // no matter what, we want to restore the handler that was changed.
       this.setSpecialMessageHandler('_rpc_applicationConnected:', null, applicationConnectedHandler);
     });
+  }
+
+  isValidPageDict(pageDict) {
+    if (!pageDict) {
+      return false;
+    }
+    let minimumInvalidPageIdentifierKey = 5; // see https://github.com/appium/appium-remote-debugger/issues/45
+    for (let pageDictValue of _.values(pageDict)) {
+      if (pageDictValue.WIRPageIdentifierKey >= minimumInvalidPageIdentifierKey) {
+        return false;
+      }
+    }
+    return true;
   }
 
   async send (command, opts = {}) {


jlipps:
@cschabetsberger-micro-focus thanks for this change! I like the idea. Do you have any idea about that null property error on Travis, thought? Also, please ensure the CLA is signed.

cschabetsberger-micro-focus:
@jlipps I tried to find out why they are failing but so far I haven't found the reason. Even if I revert my changes I am getting the same error, that plist[0] is null in test\helpers\remote-debugger-server.js. The data package that is received looks like the following: [null]-_-_-_-_|   /\_/\

Even if I explicitly check if the parsed plist item is correct, I am getting errors during test execution:
...\node_modules\bplist-parser\bplistParser.js:78
    throw new Error("maxObjectCount exceeded");
          ^-_-_-_-_-_-_-_-_-_-_-_-  ""  ""
Error: maxObjectCount exceeded
    at Object.exports.parseBuffer (...\node_modules\bplist-parser\bplistParser.js:78:11)
    at Socket.<anonymous> (..\..\../test/helpers/remote-debugger-server.js:282:35)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:188:7)
    at readableAddChunk (_stream_readable.js:176:18)
    at Socket.Readable.push (_stream_readable.js:134:10)
    at TCP.onread (net.js:548:20)

Can somebody with more knowledge about these tests have a quick look here please? Because I am somehow stuck with this.

lib/remote-debugger-rpc-client.js
```diff
@@ -141,6 +146,19 @@ export default class RemoteDebuggerRpcClient {
     });
   }
 
+  isValidPageDict(pageDict) {
+    if (!pageDict) {
+      return false;
+    }
+    let minimumInvalidPageIdentifierKey = 5; // see https://github.com/appium/appium-remote-debugger/issues/45
```
imurchie: @jlipps Do you know where this issue was moved to when the issues were all brought in from the various repos? It is no longer found, so I cannot see what this is all about.

lib/remote-debugger-rpc-client.js
```diff
@@ -141,6 +146,19 @@ export default class RemoteDebuggerRpcClient {
...
+    for (let pageDictValue of _.values(pageDict)) {
+      if (pageDictValue.WIRPageIdentifierKey >= minimumInvalidPageIdentifierKey) {
```
imurchie: I don't really understand what is going on here. To my mind, "minimum" means that it should be higher than that, but this is saying it is invalid if it is greater than the minimum.

cschabetsberger-micro-focus: Sorry for my late response. I did at lot more testing and found out that the fix in this pull request does not really fix the issue. With the fix I got iOS 10 web tests stable in my environment but not when using it on iOS 9.3 for example and also not when using hybrid apps for testing. I will try to come put with another approach to fix this.

jlipps:
OK, let's close this for now in that case. Thanks for the investigation @cschabetsberger-micro-focus

PR #55, opened 2017-01-13
filter out pages empty url for real devices. 

yaphatak:
This is already done for simulator. Applying same for real devices

### DIFF
--- a/lib/webkit-remote-debugger.js
+++ b/lib/webkit-remote-debugger.js
@@ -33,7 +33,7 @@ export default class WebKitRemoteDebugger extends RemoteDebugger {
     return !!(this.rpcClient && this.rpcClient.isConnected());
   }
 
-  async pageArrayFromJson () {
+  async pageArrayFromJson (ignoreAboutBlankUrl = false) {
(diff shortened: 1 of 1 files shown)


lib/webkit-remote-debugger.js
```diff
@@ -52,7 +52,12 @@ export default class WebKitRemoteDebugger extends RemoteDebugger {
     log.debug(`Page element JSON: ${simpleStringify(pageElementJSON)}`);
 
     // Add elements to an array
-    let newPageArray = pageElementJSON.map((pageObject) => {
+    let newPageArray = pageElementJSON.filter((pageObject) => {
+      if (pageObject.url && (!ignoreAboutBlankUrl || pageObject.url !== 'about:blank')) {
```
imurchie: This can be just `return pageObject.url && (!ignoreAboutBlankUrl || pageObject.url !== 'about:blank')`.

imurchie:
Thanks!

imurchie:
Published in `appium-remote-debugger@3.2.3`

PR #58, opened 2017-06-05
Allow alert handler to be passed in to url nav function

imurchie:
1. Add timeout for page readiness check (and option to pass in the timeout, as `pageReadyTimeout` (in ms)).
2. Add optional async function parameter to `navToUrl` to handle any possible popup during url navigation. What this does is up to the calling code.


imurchie:
Good idea. Changed.

PR #61, opened 2017-09-22
Add handler for starting/stopping JS console capture

imurchie:
Add methods `startConsole` and `stopConsole` to begin and end capture of the JS console from the connected app.

The callback sent to `startConsole` is passed an entry whenever a log line is generated on the device. The entry for something like `console.log('HI ALL');` on the device would be:
```
{ source: 'console-api',
  level: 'log',
  text: 'HI ALL',
  type: 'log',
  line: 8,
  column: 16,
  url: 'http://localhost:4994/test/guinea-pig',
  repeatCount: 1,
  parameters: [ { type: 'string', value: 'HI ALL' } ],
  stackTrace:
   [ { functionName: 'global code',
       url: 'http://localhost:4994/test/guinea-pig',
       scriptId: '1',
       lineNumber: 8,
       columnNumber: 16 } ] }
```

### DIFF
--- a/lib/remote-debugger-message-handler.js
+++ b/lib/remote-debugger-message-handler.js
@@ -128,9 +128,9 @@ export default class RpcMessageHandler {
       }
     } else if (dataKey.method === 'Page.loadEventFired') {
       await this.specialHandlers.pageLoad();
-    } else if (dataKey.method === 'Timeline.eventRecorded') {
+    } else if (dataKey.method === 'Timeline.eventRecorded' && _.isFunction(this.timelineEventHandler)) {
       this.timelineEventHandler(dataKey.params.record);
-    } else if (dataKey.method === 'Console.messageAdded') {
+    } else if (dataKey.method === 'Console.messageAdded' && _.isFunction(this.consoleLogEventHandler)) {
       this.consoleLogEventHandler(dataKey.params.message);
     } else if (_.isFunction(this.dataHandlers[msgId])) {
       log.debug('Found data handler for response');
--- a/lib/remote-debugger-message-handler.js
+++ b/lib/remote-debugger-message-handler.js
@@ -29,6 +29,10 @@ export default class RpcMessageHandler {
     this.timelineEventHandler = timelineEventHandler;
   }
 
+  setConsoleLogEventHandler (consoleLogEventHandler) {
+    this.consoleLogEventHandler = consoleLogEventHandler;
+  }
+
   hasErrorHandler (key) {
     return _.has(this.errorHandlers, key);
   }
@@ -126,6 +130,8 @@ export default class RpcMessageHandler {
       await this.specialHandlers.pageLoad();
     } else if (dataKey.method === 'Timeline.eventRecorded') {
       this.timelineEventHandler(dataKey.params.record);
+    } else if (dataKey.method === 'Console.messageAdded') {
+      this.consoleLogEventHandler(dataKey.params.message);
     } else if (_.isFunction(this.dataHandlers[msgId])) {
       log.debug('Found data handler for response');
       // we will either get back a result object that has a result.value
--- a/lib/remote-debugger.js
+++ b/lib/remote-debugger.js
@@ -501,6 +501,25 @@ class RemoteDebugger extends events.EventEmitter {
     });
   }
 
+  async startConsole (fn) {
+    log.debug('Starting to listen for JavaScript console');
+    this.rpcClient.setConsoleLogEventHandler(fn);
+    return await this.rpcClient.send('startConsole', {
+      appIdKey: this.appIdKey,
+      pageIdKey: this.pageIdKey,
+      debuggerType: this.debuggerType
+    });
+  }
+
+  async stopConsole () {
+    log.debug('Stopping to listen for JavaScript console');
+    await this.rpcClient.send('stopConsole', {
+      appIdKey: this.appIdKey,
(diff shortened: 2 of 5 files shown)


lib/remote-debugger-message-handler.js
```diff
@@ -126,6 +130,8 @@ export default class RpcMessageHandler {
       await this.specialHandlers.pageLoad();
     } else if (dataKey.method === 'Timeline.eventRecorded') {
       this.timelineEventHandler(dataKey.params.record);
+    } else if (dataKey.method === 'Console.messageAdded') {
+      this.consoleLogEventHandler(dataKey.params.message);
```
mykola-mokhnach: should we check if the handler is set first?
also, switch operator might look better here.

lib/remote-debugger.js
```diff
@@ -501,6 +501,25 @@ class RemoteDebugger extends events.EventEmitter {
...
+    log.debug('Starting to listen for JavaScript console');
+    this.rpcClient.setConsoleLogEventHandler(fn);
+    return await this.rpcClient.send('startConsole', {
+      appIdKey: this.appIdKey,
+      pageIdKey: this.pageIdKey,
+      debuggerType: this.debuggerType
+    });
+  }
+
+  async stopConsole () {
```
mykola-mokhnach: should we check if it is already started here or this does not matter?

lib/remote-messages.js
```diff
@@ -184,6 +194,14 @@ export default function getRemoteCommand (command, opts) {
       cmd = stopTimeline(opts.connId, opts.senderId, opts.appIdKey,
               opts.pageIdKey, opts.debuggerType);
       break;
+    case 'startConsole':
+      cmd = startConsole(opts.connId, opts.senderId, opts.appIdKey,
```
mykola-mokhnach: why don't we return immediately here? This could save us several lines for these _break_ operators

lib/remote-debugger-message-handler.js
```diff
@@ -29,6 +29,10 @@ export default class RpcMessageHandler {
     this.timelineEventHandler = timelineEventHandler;
   }
 
+  setConsoleLogEventHandler (consoleLogEventHandler) {
```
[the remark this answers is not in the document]
imurchie: Not really. Perhaps in a refactor to change the rest as well but not in this PR.

lib/remote-debugger-message-handler.js
imurchie: Sure. As for the `switch` statement, it could, except it ends up being a mixed conditional (not every condition is predicated on the same factor). Also, that would go into some refactoring, rather than a feature.

lib/remote-debugger.js
imurchie: (a) it doesn't matter, and (b) there is no way to tell.

lib/remote-messages.js
imurchie: As a point of programming style, I would much rather a bunch of `break` statements and a single point of exit than a whole bunch of `return` statements, even if it is longer.

imurchie:
Yeah. We've had a number of requests for Safari console log support. And watching the presentation of @christian-bromann I realized we could probably do it. And we can!

lib/remote-debugger-message-handler.js
```diff
@@ -124,8 +128,10 @@ export default class RpcMessageHandler {
       }
     } else if (dataKey.method === 'Page.loadEventFired') {
       await this.specialHandlers.pageLoad();
-    } else if (dataKey.method === 'Timeline.eventRecorded') {
+    } else if (dataKey.method === 'Timeline.eventRecorded' && _.isFunction(this.timelineEventHandler)) {
       this.timelineEventHandler(dataKey.params.record);
+    } else if (dataKey.method === 'Console.messageAdded' && _.isFunction(this.consoleLogEventHandler)) {
```
christian-bromann: @imurchie this is not the only CDP method to capture logs. Depending on the Chrome version the following methods are used:

- Runtime.consoleAPICalled
- Log.entryAdded
- Console.messageAdded

Also it might also be interesting for users to add error messages to this. For that you would need to capture data from `Runtime.exceptionThrown`.

lib/remote-messages.js
```diff
@@ -86,6 +86,16 @@ function stopTimeline (connId, senderId, appIdKey, pageIdKey, debuggerType) {
                          pageIdKey, debuggerType);
 }
 
+function startConsole (connId, senderId, appIdKey, pageIdKey, debuggerType) {
+  return command('Console.enable', {}, appIdKey, connId, senderId,
```
christian-bromann: Again, this won't work for newer Chrome versions. In fact it is a depcreated domain in the CDP protocol: https://chromedevtools.github.io/devtools-protocol/tot/Console/

imurchie: Yes. But we only use this for Safari and Safari doesn't support the new domain.

lib/remote-debugger-message-handler.js
imurchie: Interesting. This is the only one that I could get Safari to signal. I'll try harder to get others.

christian-bromann: I am not entirely sure if this works for safari since the protocol has diverged quite a lot with the recent chrome versions. I just know that it works for Chrome.

imurchie: It's worth a shot, though! Thanks for the input!

PR #63, opened 2018-02-08
Clean up logging of connection process

imurchie:
Make it a little cleaner and more informative.


imurchie:
Published in `appium-remote-debugger@3.8.1`.

PR #65, opened 2018-03-08
Add ability to communicate over unix domain sockets

imurchie:


lib/remote-debugger-rpc-client.js
```diff
@@ -14,11 +14,15 @@ import { simpleStringify } from './helpers';
 
 
 export default class RemoteDebuggerRpcClient {
-  constructor (host, port = REMOTE_DEBUGGER_PORT, specialMessageHandlers = {}) {
-    // socket and connection configuration
+  constructor (opts = {}) {
+    let {host, port, socketPath, specialMessageHandlers} = opts;
```
mykola-mokhnach: you can set default values right here

lib/remote-debugger-rpc-client.js
```diff
@@ -14,11 +14,15 @@ import { simpleStringify } from './helpers';
 
 
 export default class RemoteDebuggerRpcClient {
-  constructor (host, port = REMOTE_DEBUGGER_PORT, specialMessageHandlers = {}) {
-    // socket and connection configuration
+  constructor (opts = {}) {
```
mykola-mokhnach: do we need to raise the major lib version after his change?

imurchie: No. Everything works exactly as it did unless you pass in the socket path, at which time things change.

lib/remote-debugger.js
```diff
@@ -39,22 +39,32 @@ class RemoteDebugger extends events.EventEmitter {
   constructor (opts = {}) {
     super();
 
-    let {bundleId, platformVersion, debuggerType, useNewSafari, pageLoadMs,
-         host, port, pageReadyTimeout} = opts;
+    const {
```
jlipps: ❤️ the way this pattern looks

PR #69, opened 2018-05-01
Update dependencies

imurchie:


package.json
```diff
@@ -27,15 +27,15 @@
...
     "bplist-parser": "^0.1.0",
     "bufferpack": "0.0.6",
     "lodash": "^4.2.1",
     "request": "^2.79.0",
     "request-promise": "^4.1.1",
-    "source-map-support": "^0.4.0",
+    "source-map-support": "^0.5.5",
     "uuid": "^3.0.1",
-    "ws": "^1.1.1"
+    "ws": "^5.1.1"
```
mykola-mokhnach: Is v.5 interface of this module still compatible to v.1?

imurchie: Certainly seems to, at least in the usage `appium-remote-debugger` makes of it. I just linked this change, and ran the whole `appium-xcuitest-driver` web test suite against a real device, with no errors.

PR #72, opened 2018-05-22
Add webkit handler for Console.enable event

imurchie:
iOS 11.3 real devices sometimes reply to the start console log, so handle that otherwise it hangs.


lib/webkit-rpc-client.js
```diff
@@ -160,23 +160,19 @@ export default class WebKitRpcClient extends events.EventEmitter {
...
-    this.handlers = {
-      'Runtime.evaluate': (data) => {
-        let msgId = data.id;
-        if (data.error) {
-          this.errorHandlers[msgId](data.error);
-        }
+    const genericHandler = (data) => {
+      let msgId = data.id;
+      if (data.error) {
+        this.errorHandlers[msgId](data.error);
```
mykola-mokhnach: should we return here?

PR #81, opened 2018-06-27
Add option to use a remote debugger proxy

ThomasFettSauce:
All further info regarding this pull request can be found at https://github.com/appium/appium/pull/10960

### DIFF
--- a/lib/remote-debugger-rpc-client.js
+++ b/lib/remote-debugger-rpc-client.js
@@ -50,11 +50,12 @@ export default class RemoteDebuggerRpcClient {
     if (this.socketPath) {
       if (this.messageProxySocketPath) {
         // unix domain socket via proxy
-        log.debug(`Connecting to remote debugger through unix domain socket: '${this.messageProxySocketPath}'`);
+        log.debug(`Connecting to remote debugger via proxy through unix domain socket: '${this.messageProxySocketPath}'`);
         this.socket = net.connect(this.messageProxySocketPath);
 
         // Forward the actual socketPath to the proxy
         this.socket.once('connect', () => {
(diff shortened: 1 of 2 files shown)


lib/remote-debugger-rpc-client.js
```diff
@@ -44,10 +48,27 @@ export default class RemoteDebuggerRpcClient {
 
     // create socket and handle its messages
     if (this.socketPath) {
-      // unix domain socket
-      log.debug(`Connecting to remote debugger through unix domain socket: '${this.socketPath}'`);
-      this.socket = net.connect(this.socketPath);
+      if (this.messageProxySocketPath) {
+        // unix domain socket via proxy
+        log.debug(`Connecting to remote debugger through unix domain socket: '${this.messageProxySocketPath}'`);
```
imurchie: I think there should be some indication in the logs, here and below, that this is being proxied.

lib/remote-debugger-rpc-client.js
```diff
@@ -44,10 +48,27 @@ export default class RemoteDebuggerRpcClient {
...
+        this.socket = net.connect(this.messageProxySocketPath);
+
+        // Forward the actual socketPath to the proxy
+        this.socket.once('connect', () => {
```
mykola-mokhnach: shouldn't the event handler be assigned before calling connect?

imurchie: That is not possible. The socket is created by `net.connect`.

lib/remote-debugger-rpc-client.js
ThomasFettSauce: Good point! I adjusted the logs.

imurchie:
Published in `appium-remote-debugger@3.14.0`.

PR #84, opened 2018-07-06
Add methods for getting the cookies and delete a particular cookie

imurchie:
Safari uses the deprecated [Page.getCookies](https://chromedevtools.github.io/devtools-protocol/tot/Page#method-getCookies) and [Page.deleteCookie](https://chromedevtools.github.io/devtools-protocol/tot/Page#method-deleteCookie) methods, rather than those from the [Network domain](https://chromedevtools.github.io/devtools-protocol/tot/Network). This means that it also does not have access to setting cookies.

### DIFF
--- a/lib/remote-debugger-message-handler.js
+++ b/lib/remote-debugger-message-handler.js
@@ -85,8 +85,8 @@ export default class RpcMessageHandler {
   }
 
   async handleDataMessage (plist) {
-    let dataKey = JSON.parse(plist.__argument.WIRMessageDataKey.toString('utf8'));
-    let msgId = dataKey.id;
+    const dataKey = JSON.parse(plist.__argument.WIRMessageDataKey.toString('utf8'));
+    const msgId = (dataKey.id || '').toString();
     let result = dataKey.result;
     let error = dataKey.error || null;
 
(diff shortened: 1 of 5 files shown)


lib/remote-debugger-message-handler.js
```diff
@@ -85,8 +85,8 @@ export default class RpcMessageHandler {
   }
 
   async handleDataMessage (plist) {
-    let dataKey = JSON.parse(plist.__argument.WIRMessageDataKey.toString('utf8'));
-    let msgId = dataKey.id;
+    const dataKey = JSON.parse(plist.__argument.WIRMessageDataKey.toString('utf8'));
```
mykola-mokhnach: shall we enclose this block into try/catch or the structure of `plist` argument is a reliable thing?

lib/remote-debugger-message-handler.js
```diff
@@ -98,9 +98,6 @@ export default class RpcMessageHandler {
       error = new Error(message);
     }
 
-    if (!_.isNull(msgId) && !_.isUndefined(msgId)) {
```
mykola-mokhnach: why this is not needed?

lib/remote-debugger-message-handler.js
imurchie: To this point, this has been reliable.

mykola-mokhnach: I'd rather apply safer approach here and show the actual content of `plist` argument if something fails in this expression. But I believe you know better.

imurchie: Ok. Can't hurt.

imurchie:
I guess, with a truncate to make sure it's not super long.

PR #89, opened 2018-07-27
combine remoteDebugProxyPort and remoteDebugProxySocketPath into remo…

yfangsl:
…teDebugProxy

More info at appium/appium#10960 and the previous related PR appium/appium-remote-debugger#81.

This is based on a comment in the former to just have one cap


imurchie:
Published in `appium-remote-debugger@3.19.0` (but currently npm is having troubles with publishes, so it is not showing up).

PR #90, opened 2018-07-31
Change the way errors are handled in catch

mykola-mokhnach:
Perhaps bluebird cannot properly do its job, so we try to distinguish the actual error class it on our own


imurchie:
Published in `appium-remote-debugger@3.19.1`.

PR #91, opened 2018-08-01
Properly catch warnings inside rejectCall

mykola-mokhnach:
https://github.com/appium/appium/issues/11131


imurchie:
Published in `appium-remote-debugger@3.19.2`.

PR #92, opened 2018-08-02
Emit event when frame detaches

imurchie:
Some minor changes. The substantive change is to emit an event when frames detach. This will allow us to void the frames in the drivers.

### DIFF
--- a/lib/remote-debugger.js
+++ b/lib/remote-debugger.js
@@ -252,7 +252,15 @@ class RemoteDebugger extends events.EventEmitter {
             // if the page dictionary has not been loaded yet from the web
             // inspector, wait for it or time out after 10s
             if (appDict.pageDict.promise) {
-              await Promise.any([Promise.delay(10000), appDict.pageDict.promise]);
+              try {
+                Promise.resolve(appDict.pageDict.promise).timeout(10000);
+              } catch (err) {
+                if (!(err instanceof Promise.TimeoutError)) {
+                  throw err;
+                }
+                // on timeout, just move on
+                continue;
+              }
             }
 
             for (const dict of (appDict.pageDict || [])) {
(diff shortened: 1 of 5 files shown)


imurchie:
I am rethinking frame handling in general, but this will fix things in the interim.

lib/remote-debugger.js
```diff
@@ -245,9 +246,16 @@ class RemoteDebugger extends events.EventEmitter {
...
           let found = false;
-          dictLoop: for (let appDict of _.values(this.appDict)) {
+          dictLoop: for (const appDict of _.values(this.appDict)) {
             if (found) break; // eslint-disable-line curly
-            for (let dict of (appDict.pageDict || [])) {
+
+            // if the page dictionary has not been loaded yet from the web
+            // inspector, wait for it or time out after 10s
+            if (appDict.pageDict.promise) {
+              await Promise.any([Promise.delay(10000), appDict.pageDict.promise]);
```
mykola-mokhnach: shall we ignore the result if delay wins the race?

mykola-mokhnach: we could also do something like `await Promise.resolve(appDict.pageDict.promise).timeout(...)`

lib/remote-debugger.js
```diff
@@ -245,9 +246,24 @@ class RemoteDebugger extends events.EventEmitter {
...
             if (found) break; // eslint-disable-line curly
-            for (let dict of (appDict.pageDict || [])) {
+
+            // if the page dictionary has not been loaded yet from the web
+            // inspector, wait for it or time out after 10s
+            if (appDict.pageDict.promise) {
+              try {
+                await Promise.resolve(appDict.pageDict.promise).timeout(10000);
+              } catch (err) {
+                if (!(err instanceof Promise.TimeoutError)) {
```
mykola-mokhnach: does instanceof work as expected here or it would be better to use an explicit `constructor.name` comparison?

imurchie: It does work.

PR #93, opened 2018-08-03
 [RD-31525] Add garbage collection before javascript execution 

earendil:


imurchie:
Thanks for this idea! Not sure why you closed, but I've implemented in #94.

PR #94, opened 2018-08-03
Add JS garbage collection

imurchie:
Modeled after #93, which was closed for some reason. Add an option, `garbageCollectOnExecute`, to turn off.

### DIFF
--- a/lib/remote-debugger.js
+++ b/lib/remote-debugger.js
@@ -51,6 +51,7 @@ class RemoteDebugger extends events.EventEmitter {
       socketPath,
       pageReadyTimeout = PAGE_READY_TIMEOUT,
       remoteDebugProxy,
+      garbageCollectOnExecute = true,
     } = opts;
 
     this.bundleId = bundleId;
@@ -61,6 +62,7 @@ class RemoteDebugger extends events.EventEmitter {
       this.pageLoadMs = pageLoadMs;
       log.debug(`useNewSafari --> ${this.useNewSafari}`);
     }
+    this.garbageCollectOnExecute = garbageCollectOnExecute;
 
     this.host = host;
     this.port = port;
@@ -592,7 +594,9 @@ class RemoteDebugger extends events.EventEmitter {
       if (errors) throw new Error(errors); // eslint-disable-line curly
     }
 
-    await this.garbageCollect();
+    if (this.garbageCollectOnExecute) {
+      await this.garbageCollect();
+    }
 
     log.debug(`Sending javascript command ${_.truncate(command, {length: 50})}`);
     let res = await this.rpcClient.send('sendJSCommand', {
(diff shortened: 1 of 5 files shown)


lib/remote-debugger.js
```diff
@@ -666,6 +678,27 @@ class RemoteDebugger extends events.EventEmitter {
...
   }
+
+  async garbageCollect () {
+    log.debug(`Garbage collecting`);
+    const errors = checkParams({
+      appIdKey: this.appIdKey,
+      pageIdKey: this.pageIdKey,
+    });
+    if (errors) {
+      log.debug(`Unable to collect garbage at this time`);
```
mykola-mokhnach: shouldn't we return here?

imurchie: Whoops!

test/helpers/remote-debugger-server.js
```diff
@@ -164,24 +164,24 @@ class RemoteDebuggerServer {
     let plistData = JSON.parse(plist.__argument.WIRSocketDataKey.toString('utf8'));
 
     let result = {};
-    if (this.dataResponseError) {
+    if (this.dataResponseValue.length > 0) {
```
mykola-mokhnach: shall we check that this is an array first?

imurchie: It's always an array. It's assigned in the constructor and never assigned to again.

mykola-mokhnach: so the idea behind this condition is to check if the array has zero or more items?

mykola-mokhnach: I'd say it looks kind'a weird then

imurchie: Just more than zero. The test adds data it wants back, and this consumes it.

mykola-mokhnach: how about `!_.isEmpty` ? it should be safer

imurchie: There is _no_ safety issue here. If it is not an array then the tests will fail, and they will fail because someone broke the test infrastructure, which is something I would rather have some visibility into.

This code has absolutely no effect on package usage.

feltech:
Just to note pull request #93, which this was based on, was closed because the pull request was intended for our fork. We were yet to fully test the fix, so didn't intend to pull request upstream yet.

Unfortunately it didn't resolve the problem we were trying to fix.  It appeared to at first, but the problem is sporadic and cropped up again in production, even after we had applied garbage collection before every command.

Specifically, sometimes talking to the remote debugger on the iOS simulator can take several minutes to get a response, even for simple requests, on a particularly heavy website.

We ended up giving up and just using native actions as much as possible to circumvent the remote debugger.

PR #96, opened 2018-08-18
Add the default timeout to garbage collection

mykola-mokhnach:
It looks like sometimes the garbage collection operation freezes for too long. Lets make timeout to happen if this situation occurs.

### DIFF
--- a/lib/remote-debugger.js
+++ b/lib/remote-debugger.js
@@ -601,7 +601,7 @@ class RemoteDebugger extends events.EventEmitter {
     }
 
     if (this.garbageCollectOnExecute) {
-      await this.garbageCollect(GARBAGE_COLLECT_TIMEOUT);
+      await this.garbageCollect();
     }
 
     log.debug(`Sending javascript command ${_.truncate(command, {length: 50})}`);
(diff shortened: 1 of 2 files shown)


lib/remote-debugger.js
```diff
@@ -599,7 +601,7 @@ class RemoteDebugger extends events.EventEmitter {
     }
 
     if (this.garbageCollectOnExecute) {
-      await this.garbageCollect();
+      await this.garbageCollect(GARBAGE_COLLECT_TIMEOUT);
```
imurchie: Why not just put the constant as the default value, and leave it out here?

lib/remote-debugger.js
```diff
@@ -683,8 +685,8 @@ class RemoteDebugger extends events.EventEmitter {
     });
   }
 
-  async garbageCollect () {
-    log.debug(`Garbage collecting`);
+  async garbageCollect (timeoutMs = 60000) {
```
imurchie: `async garbageCollect (timeoutMs = GARBAGE_COLLECT_TIMEOUT) {`?

lib/remote-debugger.js
mykola-mokhnach: fixed

PR #110, opened 2018-10-31
Not sure why greenkeeper is not doing this

imurchie:


mykola-mokhnach:
> Not sure why greenkeeper is not doing this
Perhaps it requires some additional config?

PR #118, opened 2019-03-14
Add support for iOS 12.2

imurchie:
**The problem:** In iOS 12.2 the remote debugger stopped working, because instead of using the entire protocol, all communication is now wrapped in `Target.sendMessageToTarget` commands and `Target.dispatchMessageFromTarget` messages (which prior to iOS 12.2 were not known at all by Safari).

This PR implements, for iOS 12.2, this protocol change.

As I worked on this I also cleaned some stuff up, which muddled the diffs a bit, for which I apologize.

Confirmation that this works through real functional tests (with some necessary changes at the driver level):
- iOS driver: https://travis-ci.org/appium/appium-ios-driver/builds/504089881
- XCUITest driver: https://travis-ci.org/appium/appium-xcuitest-driver/builds/505734791

This works for simulators only. Beginning work on real device integration.

Re: https://github.com/appium/appium/issues/12239 and https://github.com/appium/appium/issues/12240

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -2,7 +2,7 @@ import log from './logger';
 import getAtom from './atoms';
 import _ from 'lodash';
 import assert from 'assert';
-import Promise from 'bluebird';
+import B from 'bluebird';
 
 
 const WEB_CONTENT_BUNDLE_ID = 'com.apple.WebKit.WebContent';
@@ -39,7 +39,7 @@ function pageArrayFromDict (pageDict) {
     return [pageDict];
   }
   let newPageArray = [];
-  for (let dict of _.values(pageDict)) {
+  for (const dict of _.values(pageDict)) {
     // count only WIRTypeWeb pages and ignore all others (WIRTypeJavaScript etc)
     if (_.isUndefined(dict.WIRTypeKey) || dict.WIRTypeKey === 'WIRTypeWeb') {
       newPageArray.push({
@@ -60,7 +60,7 @@ function pageArrayFromDict (pageDict) {
 function getDebuggerAppKey (bundleId, platformVersion, appDict) {
   let appId;
   if (parseFloat(platformVersion) >= 8) {
-    for (let [key, data] of _.toPairs(appDict)) {
+    for (const [key, data] of _.toPairs(appDict)) {
       if (data.bundleId === bundleId) {
         appId = key;
         break;
@@ -70,7 +70,7 @@ function getDebuggerAppKey (bundleId, platformVersion, appDict) {
     if (appId) {
       log.debug(`Found app id key '${appId}' for bundle '${bundleId}'`);
       let proxiedAppIds = [];
-      for (let [key, data] of _.toPairs(appDict)) {
+      for (const [key, data] of _.toPairs(appDict)) {
         if (data.isProxy && data.hostId === appId) {
           log.debug(`Found separate bundleId '${data.bundleId}' ` +
                     `acting as proxy for '${bundleId}', with app id '${key}'`);
@@ -94,7 +94,7 @@ function getDebuggerAppKey (bundleId, platformVersion, appDict) {
 
 function appIdForBundle (bundleId, appDict) {
   let appId;
-  for (let [key, data] of _.toPairs(appDict)) {
+  for (const [key, data] of _.toPairs(appDict)) {
     if (data.bundleId === bundleId) {
       appId = key;
       break;
@@ -112,12 +112,13 @@ function appIdForBundle (bundleId, appDict) {
 function getPossibleDebuggerAppKeys (bundleId, platformVersion, appDict) {
   let proxiedAppIds = [];
   if (parseFloat(platformVersion) >= 8) {
-    let appId = appIdForBundle(bundleId, appDict);
+    const appId = appIdForBundle(bundleId, appDict);
+    proxiedAppIds.push(appId);
 
     // now we need to determine if we should pick a proxy for this instead
     if (appId) {
       log.debug(`Found app id key '${appId}' for bundle '${bundleId}'`);
-      for (let [key, data] of _.toPairs(appDict)) {
+      for (const [key, data] of _.toPairs(appDict)) {
         if (data.isProxy && data.hostId === appId) {
           log.debug(`Found separate bundleId '${data.bundleId}' ` +
                     `acting as proxy for '${bundleId}', with app id '${key}'`);
@@ -139,7 +140,7 @@ function getPossibleDebuggerAppKeys (bundleId, platformVersion, appDict) {
 
 function checkParams (params) {
   let errors = [];
-  for (let [param, value] of _.toPairs(params)) {
+  for (const [param, value] of _.toPairs(params)) {
     try {
       assert.ok(value);
     } catch (err) {
@@ -163,7 +164,7 @@ async function getScriptForAtom (atom, args, frames, asyncCallBack = null) {
   let script;
   if (frames.length > 0) {
     script = atomSrc;
-    for (let frame of frames) {
+    for (const frame of frames) {
       script = await wrapScriptForFrame(script, frame);
     }
   } else {
@@ -190,7 +191,7 @@ function simpleStringify (value) {
   // we get back objects sometimes with string versions of functions
   // which muddy the logs
   let cleanValue = _.clone(value);
-  for (let property of ['ceil', 'clone', 'floor', 'round', 'scale', 'toString']) {
+  for (const property of ['ceil', 'clone', 'floor', 'round', 'scale', 'toString']) {
     delete cleanValue[property];
   }
   return JSON.stringify(cleanValue);
@@ -200,7 +201,7 @@ function deferredPromise () {
   // http://bluebirdjs.com/docs/api/deferred-migration.html
   let resolve;
   let reject;
-  let promise = new Promise((res, rej) => { // eslint-disable-line promise/param-names
+  const promise = new B((res, rej) => { // eslint-disable-line promise/param-names
     resolve = res;
     reject = rej;
   });
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -12,11 +12,11 @@ const WEB_CONTENT_BUNDLE_ID = 'com.apple.WebKit.WebContent';
  * dictionary whose keys are understandable
  */
 function appInfoFromDict (dict) {
-  let id = dict.WIRApplicationIdentifierKey;
-  let isProxy = _.isString(dict.WIRIsApplicationProxyKey)
+  const id = dict.WIRApplicationIdentifierKey;
+  const isProxy = _.isString(dict.WIRIsApplicationProxyKey)
     ? dict.WIRIsApplicationProxyKey.toLowerCase() === 'true'
     : dict.WIRIsApplicationProxyKey;
-  let entry = {
+  const entry = {
     id,
     isProxy,
     name: dict.WIRApplicationNameKey,
@@ -113,10 +113,10 @@ function getPossibleDebuggerAppKeys (bundleId, platformVersion, appDict) {
   let proxiedAppIds = [];
   if (parseFloat(platformVersion) >= 8) {
     const appId = appIdForBundle(bundleId, appDict);
-    proxiedAppIds.push(appId);
 
     // now we need to determine if we should pick a proxy for this instead
     if (appId) {
+      proxiedAppIds.push(appId);
       log.debug(`Found app id key '${appId}' for bundle '${bundleId}'`);
       for (const [key, data] of _.toPairs(appDict)) {
         if (data.isProxy && data.hostId === appId) {
@@ -124,9 +124,6 @@ function getPossibleDebuggerAppKeys (bundleId, platformVersion, appDict) {
                     `acting as proxy for '${bundleId}', with app id '${key}'`);
           proxiedAppIds.push(key);
         }
-      }
-      if (proxiedAppIds.length === 0) {
-        proxiedAppIds = [appId];
       }
     }
   } else {
--- a/lib/message-handlers.js
+++ b/lib/message-handlers.js
@@ -1,21 +1,43 @@
 import log from './logger';
 import { RemoteDebugger } from './remote-debugger';
 import { pageArrayFromDict, getDebuggerAppKey, simpleStringify } from './helpers';
+import _ from 'lodash';
+
 
 /*
  * Generic callbacks used throughout the lifecycle of the Remote Debugger.
  * These will be added to the prototype.
  */
 
+
+/**
+ * Remove the `isKey` property from the page array, since it does not affect
+ * equality
+ */
+function cleanPageArray (arr) {
+  return _.map(arr, (el) => _.pick(el, 'id', 'title', 'url'));
+}
+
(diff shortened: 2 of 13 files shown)


lib/helpers.js
```diff
@@ -112,12 +112,13 @@ function appIdForBundle (bundleId, appDict) {
 function getPossibleDebuggerAppKeys (bundleId, platformVersion, appDict) {
   let proxiedAppIds = [];
   if (parseFloat(platformVersion) >= 8) {
-    let appId = appIdForBundle(bundleId, appDict);
+    const appId = appIdForBundle(bundleId, appDict);
+    proxiedAppIds.push(appId);
```
mykola-mokhnach: is this needed?

lib/message-handlers.js
```diff
@@ -1,21 +1,43 @@
...
-    if (this.appDict[appIdKey].pageDict && this.appDict[appIdKey].pageDict.resolve) {
-      // pageDict is a promise, so resolve
-      this.appDict[appIdKey].pageDict.resolve(pageDict);
+    if (this.appDict[appIdKey].pageDict) {
+      if (this.appDict[appIdKey].pageDict.resolve) {
+        // pageDict is a promise, so resolve
+        this.appDict[appIdKey].pageDict.resolve(pageArray);
+      } else {
+        // we have a pre-existing pageDict
+        if (_.isEqual(cleanPageArray(this.appDict[appIdKey].pageDict), cleanPageArray(pageArray))) {
```
mykola-mokhnach: can we pass a dictionary to cleanPageArray?

lib/message-handlers.js
```diff
@@ -1,21 +1,43 @@
...
+
   // save the page dict for this app
   if (this.appDict[appIdKey]) {
-    if (this.appDict[appIdKey].pageDict && this.appDict[appIdKey].pageDict.resolve) {
-      // pageDict is a promise, so resolve
-      this.appDict[appIdKey].pageDict.resolve(pageDict);
+    if (this.appDict[appIdKey].pageDict) {
+      if (this.appDict[appIdKey].pageDict.resolve) {
+        // pageDict is a promise, so resolve
+        this.appDict[appIdKey].pageDict.resolve(pageArray);
```
mykola-mokhnach: does this mutate pageArray?

lib/remote-debugger-message-handler.js
```diff
@@ -2,13 +2,26 @@ import log from './logger';
...
+  'Page.frameStartedLoading',
+  'Page.frameStoppedLoading',
+  'Page.frameScheduledNavigation',
+  'Page.frameClearedScheduledNavigation',
+  'Console.messagesCleared',
+];
+
 export default class RpcMessageHandler {
-  constructor (specialHandlers) {
+  constructor (specialHandlers, targetBased = false) {
```
mykola-mokhnach: isTargetBased?

lib/remote-debugger-message-handler.js
```diff
@@ -49,22 +62,22 @@ export default class RpcMessageHandler {
     this.willNavigateWithoutReload = allow;
   }
 
-  handleMessage (plist) {
-    let handlerFor = plist.__selector;
-    if (!handlerFor) {
+  async handleMessage (plist) {
```
mykola-mokhnach: have you verified this method is called with `await` everywhere now?

lib/remote-debugger-message-handler.js
```diff
@@ -120,10 +133,45 @@ export default class RpcMessageHandler {
       return;
     }
 
-    if (dataKey.method === 'Profiler.resetProfiles') {
+    let method = dataKey.method;
+    let params;
+    if (this.targetBased) {
+      if (method === 'Target.targetCreated') {
```
mykola-mokhnach: I'd rather move this code into a separate method. The current one is already quite long

lib/remote-debugger-message-handler.js
```diff
@@ -120,10 +133,45 @@ export default class RpcMessageHandler {
...
+        await this.specialHandlers.targetDestroyed(app, targetInfo);
+        return;
+      } else if (dataKey.method !== 'Target.dispatchMessageFromTarget') {
+        if (!_.isEmpty(msgId)) {
+          log.debug(`Received receipt for message '${msgId}'`);
+        }
+        return;
+      }
+
+      const message = JSON.parse(dataKey.params.message);
```
mykola-mokhnach: this might throw an exception

lib/remote-debugger-message-handler.js
```diff
@@ -120,10 +133,45 @@ export default class RpcMessageHandler {
...
+      result = message.result || message;
+      msgId = message.id;
+      method = message.method;
+      params = result.params;
```
mykola-mokhnach: can result be non-object?

lib/remote-debugger-message-handler.js
```diff
@@ -120,10 +133,45 @@ export default class RpcMessageHandler {
...
+      } else if (dataKey.method !== 'Target.dispatchMessageFromTarget') {
+        if (!_.isEmpty(msgId)) {
+          log.debug(`Received receipt for message '${msgId}'`);
+        }
+        return;
+      }
+
+      const message = JSON.parse(dataKey.params.message);
+      result = message.result || message;
+      msgId = message.id;
```
mykola-mokhnach: can message be not an object?

lib/remote-debugger-rpc-client.js
```diff
@@ -3,17 +3,20 @@ import _ from 'lodash';
...
+import RemoteMessages from './remote-messages';
 
 
+const MIN_PLATFORM_FOR_TARGET_BASED = 12.2;
+
 export default class RemoteDebuggerRpcClient {
   constructor (opts = {}) {
-    const {
+    let {
+      platformVersion = {},
```
mykola-mokhnach: we are currently not getting xcode version for real devices, because it might not be installed on the target machine. Do we want to change that?

lib/remote-debugger-rpc-client.js
```diff
@@ -37,12 +40,17 @@ export default class RemoteDebuggerRpcClient {
 
     // message handlers
     this.specialMessageHandlers = specialMessageHandlers;
-    this.messageHandler = null;
+
+    // on iOS 12.2 the messages get sent through the Target domain
+    if (_.isString(platformVersion)) {
+      platformVersion = parseFloat(platformVersion);
+    }
+    const targetBased = platformVersion >= MIN_PLATFORM_FOR_TARGET_BASED;
```
mykola-mokhnach: can we log if targetBased is enabled here?

lib/remote-debugger-rpc-client.js
```diff
@@ -179,17 +187,20 @@ export default class RemoteDebuggerRpcClient {
...
     // error listener, which needs to be removed after the promise is resolved
     let onSocketError;
 
-    return new Promise((resolve, reject) => {
+    return await new B((resolve, reject) => {
       // promise to be resolved whenever remote debugger
       // replies to our request
 
+      // keep track of the messages coming and going using a simple sequential id
+      const msgId = this.curMsgId++;
```
mykola-mokhnach: do we want to increase this.curMsgId after assigning its value to msgId?

lib/remote-debugger-rpc-client.js
```diff
@@ -205,53 +216,60 @@ export default class RemoteDebuggerRpcClient {
...
+        this.setDataMessageHandler(msgId.toString(), errorHandler, function (value) {
           const msg = _.truncate(_.isString(value) ? value : JSON.stringify(value), {length: 50});
           log.debug(`Received data response from socket send: '${msg}'`);
           log.debug(`Original command: ${command}`);
           resolve(value);
         });
-        data.__argument.WIRSocketDataKey.id = this.curMsgId;
-        data.__argument.WIRSocketDataKey =
-            Buffer.from(JSON.stringify(data.__argument.WIRSocketDataKey));
+        if (cmd.__argument.WIRSocketDataKey.params) {
```
mykola-mokhnach: hell it's full of magic

lib/remote-debugger.js
```diff
@@ -223,6 +208,26 @@ class RemoteDebugger extends events.EventEmitter {
...
+      if (key === 'pageDict' && !_.isArray(value)) {
+        return `'Waiting for data'`;
+      }
+      return JSON.stringify(value);
+    }
+    log.debug('Current applications available:');
+    for (const [app, info] of _.toPairs(apps)) {
+      log.debug(`    Application: '${app}'`);
+      for (const [key, value] of _.toPairs(info)) {
+        let valueString = getValueString(key, value);
```
mykola-mokhnach: const

lib/remote-debugger.js
```diff
@@ -317,18 +322,20 @@ class RemoteDebugger extends events.EventEmitter {
...
       log.debug(`Waiting for ${pagePromises.length} pages to be fulfilled`);
-      await Promise.any([Promise.delay(30000), Promise.all(pagePromises)]);
+      await B.any([B.delay(30000), B.all(pagePromises)]);
     }
 
     // translate the dictionary into a useful form, and return to sender
-    let pageArray = pageArrayFromDict(pageDict);
+    const pageArray = _.isEmpty(this.appDict[appIdKey].pageDict)
+      ? pageArrayFromDict(pageDict)
+      : this.appDict[appIdKey].pageDict;
```
mykola-mokhnach: it's a bit unclear here how pageDict is getting to an array without any transformations

lib/remote-messages.js
```diff
@@ -3,253 +3,294 @@
...
+      },
+      __selector: '_rpc_forwardSocketSetup:'
+    };
+  }
 
-function setUrl (connId, senderId, appIdKey, pageIdKey, debuggerType, url) {
-  return command('Page.navigate', {url}, appIdKey, connId,
-      senderId, pageIdKey, debuggerType);
-}
+  indicateWebView (connId, appIdKey, pageIdKey, opts) {
```
mykola-mokhnach: opts = {}

lib/remote-messages.js
```diff
@@ -3,253 +3,294 @@
...
-          includeCommandLineAPI: true,
-          doNotPauseOnExceptionsAndMuteConsole: true,
-        }
+
+  /*
+   * Internal functions
+   */
+
+  command (method, params, appIdKey, connId, senderId, pageIdKey, debuggerType) {
+    if (debuggerType !== null && debuggerType === DEBUGGER_TYPES.webkit) {
```
mykola-mokhnach: debuggerType !== null might be skipped

lib/remote-messages.js
```diff
@@ -3,253 +3,294 @@
...
+      return this.commandWebKit(method, params);
+    } else {
```
mykola-mokhnach: else is redundant

lib/remote-messages.js
```diff
@@ -3,253 +3,294 @@
...
-    case 'garbageCollect':
-      cmd = garbageCollect(opts.connId, opts.senderId, opts.appIdKey,
-              opts.pageIdKey, opts.debuggerType);
-      break;
-    default:
-      throw new Error(`Unknown command: ${command}`);
-  }
-
-  return cmd;
+  getRemoteCommand (command, opts) {
```
mykola-mokhnach: opts = {}

lib/remote-messages.js
```diff
@@ -3,253 +3,294 @@
...
+      case 'connectToApp':
+        cmd = this.connectToApp(connId, appIdKey);
+        break;
+      case 'setSenderKey':
+        cmd = this.setSenderKey(connId, senderId, appIdKey, pageIdKey);
+        break;
+      case 'indicateWebView':
+        cmd = this.indicateWebView(connId, appIdKey, pageIdKey, opts);
+        break;
+      case 'sendJSCommand':
```
mykola-mokhnach: cannot we simply call this[command](...) for all the other commands with equal arguments?

lib/message-handlers.js
imurchie: No. It doesn't actually do _anything_ to, or with it. I'll remove.

lib/message-handlers.js
imurchie: You can. Mostly because `this.appDict[appIdKey].pageDict)` is actually an array. I'll change that throughout.

lib/remote-debugger-rpc-client.js
imurchie: We do not want to change that. It limits how iOS real device testing can be done.

But in any case we _will_ have a `platformVersion`, which is what we depend on (since iOS below 12.2, but still Xcode 10.2, continues to use the old mechanism).

mykola-mokhnach:
it might be a good idea to add iOS 12.2 to Travis

imurchie:
The build for this package does not actually depend on Xcode at all. It is entirely through a test server we built. This should be changed, but it outside the scope of this work (mostly because it is a huge task).

lib/message-handlers.js
```diff
@@ -23,21 +23,21 @@ function onPageChange (appIdKey, pageDict) {
 
   // save the page dict for this app
   if (this.appDict[appIdKey]) {
-    if (this.appDict[appIdKey].pageDict) {
-      if (this.appDict[appIdKey].pageDict.resolve) {
-        // pageDict is a promise, so resolve
-        this.appDict[appIdKey].pageDict.resolve(pageArray);
+    if (this.appDict[appIdKey].pageArray) {
+      if (this.appDict[appIdKey].pageArray.resolve) {
```
mykola-mokhnach: isFunction ?

lib/remote-debugger-message-handler.js
```diff
@@ -107,6 +107,70 @@ export default class RpcMessageHandler {
...
+    } else if (_.isFunction(this.dataHandlers[msgId])) {
+      log.debug('Found data handler for response');
+
+      // we will either get back a result object that has a result.value
+      // in which case that is what we want,
+      // or else we return the whole thing
+      if (result.result && result.result.value) {
+        result = result.result.value;
+      }
+      this.dataHandlers[msgId](result);
```
mykola-mokhnach: await?

lib/remote-debugger-message-handler.js
```diff
@@ -107,6 +107,70 @@ export default class RpcMessageHandler {
...
+                  `result: '${JSON.stringify(result)}'; ` +
+                  `error: '${error}'`);
+      }
+    }
+  }
+
+  logFullMessage (plist) {
+    // Buffers cannot be serialized in a readable way
+    const bufferToJSON = Buffer.prototype.toJSON;
+    delete Buffer.prototype.toJSON;
```
mykola-mokhnach: interesting hack %)

lib/message-handlers.js
```diff
@@ -23,21 +23,21 @@ function onPageChange (appIdKey, pageDict) {
...
+        // pageDict is a pending promise, so resolve
+        this.appDict[appIdKey].pageArray.resolve();
```
mykola-mokhnach: await?

imurchie: `resolve` is just a function. The promise itself it awaited elsewhere (or has already timed out).

lib/remote-debugger-message-handler.js
imurchie: These are always regular functions.

PR #122, opened 2019-03-29
Handle iOS 12.2 for real devices

imurchie:
Add support for iOS 12.2 on real devices. For this to work `ios-webkit-debug-proxy` needs to be the latest from HEAD.

This duplicates a bunch of logic, but as the package is structured now that is necessary. Once this is in I will allocate some time to refactor this to use the same message handler system, which should remove the duplication.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -3,10 +3,11 @@ import getAtom from './atoms';
 import _ from 'lodash';
 import assert from 'assert';
 import B from 'bluebird';
+import { util } from 'appium-support';
 
 
 const WEB_CONTENT_BUNDLE_ID = 'com.apple.WebKit.WebContent';
-const MIN_PLATFORM_FOR_TARGET_BASED = 12.2;
+const MIN_PLATFORM_FOR_TARGET_BASED = '12.2';
 
 /*
  * Takes a dictionary from the remote debugger and makes a more manageable
@@ -210,16 +211,13 @@ function deferredPromise () {
   };
 }
 
-function checkIsTargetBased (isSafari, platformVersion) {
+function isTargetBased (isSafari, platformVersion) {
   // on iOS 12.2 the messages get sent through the Target domain
-  if (_.isString(platformVersion)) {
-    platformVersion = parseFloat(platformVersion);
-  }
-  return isSafari && platformVersion >= MIN_PLATFORM_FOR_TARGET_BASED;
+  return isSafari && util.compareVersions(platformVersion, '>=', MIN_PLATFORM_FOR_TARGET_BASED);
 }
 
 export {
   appInfoFromDict, pageArrayFromDict, getDebuggerAppKey,
   getPossibleDebuggerAppKeys, checkParams, wrapScriptForFrame, getScriptForAtom,
-  simpleStringify, deferredPromise, checkIsTargetBased,
+  simpleStringify, deferredPromise, isTargetBased,
 };
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -6,6 +6,7 @@ import B from 'bluebird';
 
 
 const WEB_CONTENT_BUNDLE_ID = 'com.apple.WebKit.WebContent';
+const MIN_PLATFORM_FOR_TARGET_BASED = 12.2;
 
 /*
  * Takes a dictionary from the remote debugger and makes a more manageable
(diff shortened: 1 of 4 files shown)


lib/helpers.js
```diff
@@ -209,8 +210,16 @@ function deferredPromise () {
   };
 }
 
+function checkIsTargetBased (isSafari, platformVersion) {
+  // on iOS 12.2 the messages get sent through the Target domain
+  if (_.isString(platformVersion)) {
+    platformVersion = parseFloat(platformVersion);
```
mykola-mokhnach: please use compareVersions/coerceVersion helpers from appium-support

lib/helpers.js
```diff
@@ -209,8 +210,16 @@ function deferredPromise () {
   };
 }
 
+function checkIsTargetBased (isSafari, platformVersion) {
```
mykola-mokhnach: why not isTargetBased? `check` prefix means we do some additional verification there

lib/webkit-rpc-client.js
```diff
@@ -141,16 +165,53 @@ export default class WebKitRpcClient extends events.EventEmitter {
       return rejectCall(new Error(message));
     }
 
-    const msgId = data.id;
+    let msgId = data.id;
+
+    let result = data.result;
```
mykola-mokhnach: let {result, params, error} = data

lib/webkit-rpc-client.js
```diff
@@ -141,16 +165,53 @@ export default class WebKitRpcClient extends events.EventEmitter {
...
+    let params = data.params;
+    let error = data.error;
+    let method = this.dataMethods[msgId] || data.method;
+    if (this.isTargetBased) {
+      if (!(data.method || '').startsWith('Target')) {
```
mykola-mokhnach: lodash's startsWith method could also be used here. It does safe typecast

lib/webkit-rpc-client.js
```diff
@@ -141,16 +165,53 @@ export default class WebKitRpcClient extends events.EventEmitter {
...
+        log.debug(`Received reciept for message '${msgId}'`);
+        return;
+      }
+      if (data.params.message) {
+        const message = JSON.parse(data.params.message);
```
mykola-mokhnach: this can throw an exception

lib/webkit-rpc-client.js
```diff
@@ -141,16 +165,53 @@ export default class WebKitRpcClient extends events.EventEmitter {
...
+
+        // get nested params if necessary
+        if (message.params) {
+          params = message.params;
+        }
+
+        // the message is aggravatingly nested
+        if (message.result) {
```
mykola-mokhnach: can we change that to `message.result && message.result.result && message.result.result.value` ?

lib/webkit-rpc-client.js
```diff
@@ -180,19 +245,25 @@ export default class WebKitRpcClient extends events.EventEmitter {
...
+          this.networkEventHandler(method, params);
           return;
         }
         break;
+      case 'Target.targetCreated':
+        log.debug(`Target created: ${JSON.stringify(params.targetInfo)}`);
+        isEventHandled = true;
+        break;
+      case 'Target.dispatchMessageFromTarget':
+        log.debug('HERE');
```
mykola-mokhnach: is this log useful?

lib/webkit-rpc-client.js
imurchie: No. Anything nested would end up with empty results:
```js
node
> message = {
...   "result": {
.....     "cookies": [
.....       {
.......         "name": "guineacookie1",
.......         "value": "i%20am%20a%20cookie%20value",
.......         "domain": "192.168.1.152",
.......         "path": "/",
.......         "expires": 0,
.......         "size": 40,
.......         "httpOnly": false,
.......         "secure": false,
.......         "session": true,
.......         "sameSite": "None"
.......       },
.....       {
.......         "name": "guineacookie2",
.......         "value": "cooki%C3%A92",
.......         "domain": "192.168.1.152",
.......         "path": "/",
.......         "expires": 0,
.......         "size": 25,
.......         "httpOnly": false,
.......         "secure": false,
.......         "session": true,
.......         "sameSite": "None"
.......       }
.....     ]
.....   },
...   "id": 6
... }
{ result: { cookies: [ [Object], [Object] ] }, id: 6 }
> b = message.result && message.result.result && message.result.result.value;
undefined
> b = message.result && message.result.result
undefined
> b = message.result
{ cookies:
   [ { name: 'guineacookie1',
       value: 'i%20am%20a%20cookie%20value',
       domain: '192.168.1.152',
       path: '/',
       expires: 0,
       size: 40,
       httpOnly: false,
       secure: false,
       session: true,
       sameSite: 'None' },
     { name: 'guineacookie2',
       value: 'cooki%C3%A92',
       domain: '192.168.1.152',
       path: '/',
       expires: 0,
       size: 25,
       httpOnly: false,
       secure: false,
       session: true,
       sameSite: 'None' } ] }
```

PR #125, opened 2019-04-22
Dynamically change communication protocol

imurchie:
There have been problems where the wrong communication system is used. This PR changes the logic so that the initial condition is based on what seems to be right, but then if the error comes in that the domain is not know, we switch to the other protocol.

See: 
https://github.com/appium/appium/issues/12518
https://github.com/appium/appium/issues/12510

### DIFF
--- a/lib/remote-debugger-rpc-client.js
+++ b/lib/remote-debugger-rpc-client.js
@@ -8,13 +8,15 @@ import { REMOTE_DEBUGGER_PORT } from './remote-debugger';
 import UUID from 'uuid-js';
 import net from 'net';
 import RpcMessageHandler from './remote-debugger-message-handler';
-import RemoteMessages from './remote-messages';
 import { isTargetBased } from './helpers';
+import RpcClient from './rpc-client';
 
 
-export default class RemoteDebuggerRpcClient {
+export default class RemoteDebuggerRpcClient extends RpcClient {
   constructor (opts = {}) {
-    let {
+    super();
+
+    const {
       platformVersion = {},
       isSafari = true,
       host = '::1',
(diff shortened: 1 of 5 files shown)


lib/rpc-client.js
```diff
@@ -0,0 +1,36 @@
+import events from 'events';
+import log from './logger';
+import RemoteMessages from './remote-messages';
+
+
+export default class RpcClient extends events.EventEmitter {
+  constructor () {
```
mykola-mokhnach: this is probably redundant

lib/rpc-client.js
```diff
@@ -0,0 +1,36 @@
...
+    if (!this.remoteMessages) {
+      this.remoteMessages = new RemoteMessages(isTargetBased);
+    } else {
+      this.remoteMessages.setCommunicationProtocol(isTargetBased);
+    }
+  }
+
+  async send (command, opts = {}) {
+    try {
+      return await this.sendMessage(command, opts);
```
mykola-mokhnach: should we declare an empty sendMessage method in this class to show that it must be overridden by subclasses?

lib/remote-debugger-rpc-client.js
```diff
@@ -41,10 +43,18 @@ export default class RemoteDebuggerRpcClient {
...
 
-    const targetBased = isTargetBased(isSafari, platformVersion);
-    this.remoteMessages = new RemoteMessages(targetBased);
-    this.messageHandler = new RpcMessageHandler(this.specialMessageHandlers, targetBased);
-    log.debug(`Using ${targetBased ? 'Target-based' : 'full Web Inspector protocol'} communication`);
+    // start with a best guess for the protocol
+    this.setCommunicationProtocol(isTargetBased(isSafari, platformVersion));
+  }
+
+  setCommunicationProtocol (isTargetBased = false) {
```
mykola-mokhnach: how about having setter/getter instead?

imurchie: I'm not a big fan of getters/setters that do other things.

lib/rpc-client.js
imurchie: It is if EventEmitter is removed (and I'm not sure why it is an EventEmitter).

lib/rpc-client.js
imurchie: Done!

PR #130, opened 2019-06-11
Add wait for target

imurchie:
There are cases where the device does not report (or even create, I guess) the target until after we've already sent a request, which ends up failing. So add a wait until a target has been received.

As of now I cannot find a way to map targets to any sort of marker within the app, so this is the best I can do. Perhaps this will uncover further details.

Trial run with this modification, in `appium-xcuitest-driver`: https://travis-ci.org/appium/appium-xcuitest-driver/builds/544315465

Test suite runs locally against real devices.

Working towards https://github.com/appium/appium/issues/12614


lib/helpers.js
```diff
@@ -182,7 +182,7 @@ async function getScriptForAtom (atom, args, frames, asyncCallBack = null) {
   return script;
 }
 
-function simpleStringify (value) {
+function simpleStringify (value, multiline) {
```
mykola-mokhnach: should the param has the default value?

lib/rpc-client.js
```diff
@@ -1,8 +1,20 @@
...
 import RemoteMessages from './remote-messages';
+import { retryInterval } from 'asyncbox';
+import _ from 'lodash';
 
 
+const WAIT_FOR_TARGET_RETRIES = 10;
+const WAIT_FOR_TARGET_INTERVAL = 1000;
+
 export default class RpcClient {
+  #targets;
```
mykola-mokhnach: how it works?

lib/helpers.js
imurchie: It could. It's being treated as defaulting to `false`, so absence is the same as setting a default.

lib/rpc-client.js
```diff
@@ -32,4 +58,39 @@ export default class RpcClient {
   async sendMessage (/* command, opts = {} */) { // eslint-disable-line require-await
     throw new Error(`Sub-classes need to implement a 'sendMessage' function`);
   }
+
+  addTarget (targetInfo) {
+    log.debug(`Target created: ${JSON.stringify(targetInfo)}`);
+    if (!this.targets.includes(targetInfo.targetId)) {
```
mykola-mokhnach: should we check if `this.targets` is defined?

lib/rpc-client.js
imurchie: It's a private property. I was mostly just playing with them. I don't want to actually use it because it requires locking into a lower limit on Babel, which I don't want to do.

lib/rpc-client.js
imurchie: `this.targets` is a getter, so it will always be something.

lib/helpers.js
mykola-mokhnach: I would say adding a default value will improve readability

lib/rpc-client.js
```diff
@@ -72,8 +69,8 @@ export default class RpcClient {
   }
 
   get targets () {
-    this.#targets = this.#targets || [];
-    return this.#targets;
+    this._targets = this._targets || [];
```
mykola-mokhnach: I'd set the default value in the constructor

PR #133, opened 2019-06-17
Make sure empty target events are handled correctly

imurchie:
It seems that sometimes the Web Inspector sends empty `targetDestroyed` events, which leads to errors since we're trying to get a property on the information, which would be `undefined`.

See line 10207 of https://dev.azure.com/srinivasansekar/java-client/_build/results?buildId=164

### DIFF
--- a/lib/rpc-client.js
+++ b/lib/rpc-client.js
@@ -57,14 +57,22 @@ export default class RpcClient {
     throw new Error(`Sub-classes need to implement a 'sendMessage' function`);
   }
 
-  addTarget (targetInfo = {}) {
+  addTarget (targetInfo) {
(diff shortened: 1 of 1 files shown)


lib/rpc-client.js
```diff
@@ -57,14 +57,14 @@ export default class RpcClient {
     throw new Error(`Sub-classes need to implement a 'sendMessage' function`);
   }
 
-  addTarget (targetInfo) {
+  addTarget (targetInfo = {}) {
     log.debug(`Target created: ${JSON.stringify(targetInfo)}`);
     if (!this.targets.includes(targetInfo.targetId)) {
```
mykola-mokhnach: this may put `undefined` into the `targets` list

imurchie: True... :(

PR #137, opened 2019-07-11
feat: Remove the dependency of iwdp for real devices and use appium-ios-device

umutuzgur:


lib/remote-debugger-real-device.js
```diff
@@ -0,0 +1,42 @@
...
+export default class RemoteDebuggerRealDevice extends RemoteDebugger {
+  constructor (opts = {}) {
+    super(opts);
+
+    this.udid = opts.udid;
+
+    this.skippedApps = ['lockdownd'];
+  }
+
+  async connect () {
```
mykola-mokhnach: some documentation on newly added classes and method might be really useful

lib/rpc-client-simulator.js
```diff
@@ -0,0 +1,256 @@
...
+    return await new B((resolve, reject) => {
+      // only resolve this function when we are actually connected
+      this.socket.on('connect', () => {
+        log.debug(`Debugger socket connected`);
+        this.connected = true;
+
+        resolve();
+      });
+      this.socket.on('error', (err) => {
+        if (this.connected) {
```
mykola-mokhnach: why this condition is needed?

lib/rpc-client.js
```diff
@@ -39,21 +70,98 @@ export default class RpcClient {
...
+          log.debug(`Received response from send: '${response}'`);
+
+          // call the original listener, and put it back, if necessary
+          specialMessageHandler(...args);
+          if (this.messageHandler.hasSpecialMessageHandler(cmd.__selector)) {
+            // this means that the system has not removed this listener
+            this.setSpecialMessageHandler(cmd.__selector, null, specialMessageHandler);
+          }
+
+          resolve(args);
```
mykola-mokhnach: args or `...args` ?

lib/rpc-client.js
```diff
@@ -99,4 +207,95 @@ export default class RpcClient {
...
+  }
+
+  async selectApp (appIdKey, applicationConnectedHandler) {
+    return await new B((resolve, reject) => {
+      // local callback, temporarily added as callback to
+      // `_rpc_applicationConnected:` remote debugger response
+      // to handle the initial connection
+      const onAppChange = (dict) => {
+        // from the dictionary returned, get the ids
+        let oldAppIdKey = dict.WIRHostApplicationIdentifierKey;
```
mykola-mokhnach: `const {oldAppIdKey:WIRHostApplicationIdentifierKey, correctAppIdKey:WIRApplicationIdentifierKey} = dict`

lib/remote-debugger-real-device.js
```diff
@@ -0,0 +1,42 @@
...
+import RpcClientRealDevice from './rpc-client-real-device';
+
+
+export default class RemoteDebuggerRealDevice extends RemoteDebugger {
+  constructor (opts = {}) {
+    super(opts);
+
+    this.udid = opts.udid;
+
+    this.skippedApps = ['lockdownd'];
```
mykola-mokhnach: I'd rather put this list into an overridable getter method

lib/remote-debugger-real-device.js
```diff
@@ -0,0 +1,42 @@
...
+    await this.rpcClient.connect();
+
+    // get the connection information about the app
+    try {
+      const appInfo = await this.setConnectionKey();
+      log.debug('Connected to application');
+      return appInfo;
+    } catch (err) {
+      await this.disconnect();
+      return null;
```
mykola-mokhnach: where the return result of `connect` is used?

lib/rpc-client-real-device.js
```diff
@@ -0,0 +1,40 @@
...
+import { services } from 'appium-ios-device';
+
+
+export default class RpcClientRealDevice extends RpcClient {
+  constructor (opts = {}) {
+    super(Object.assign({
+      shouldCheckForTarget: false,
+    }, opts));
+
+    const {
```
mykola-mokhnach: `this.udid = opts.udid` is enough

lib/rpc-client-real-device.js
```diff
@@ -0,0 +1,40 @@
...
+      udid,
+    } = opts;
+
+    this.udid = udid;
+  }
+
+  async connect () {
+    this.service = await services.startWebInspectorService(this.udid);
```
mykola-mokhnach: should we prevent double call in this method?

mykola-mokhnach: same comment to all `connect`mthods

lib/rpc-client-simulator.js
```diff
@@ -0,0 +1,256 @@
...
+          log.error(`Socket error: ${err.message}`);
+        }
+
+        // the connection was refused, so reject the connect promise
+        reject(err);
+      };
+      this.socket.on('error', onSocketError);
+
+      if (this.messageHandler.hasSpecialMessageHandler(cmd.__selector)) {
```
mykola-mokhnach: do we need the conditions that do nothing?

lib/rpc-client-simulator.js
```diff
@@ -0,0 +1,256 @@
...
+    let dataLeftOver = true;
+
+    // Parse multiple messages in the same packet
+    while (dataLeftOver) {
+      // Store a reference to where we were
+      const oldReadPos = this.readPos;
+
+      // Read the prefix (plist length) to see how far to read next
+      // It's always 4 bytes long
+      const prefix = this.received.slice(this.readPos, this.readPos + 4);
```
mykola-mokhnach: 4 can be a constant

lib/rpc-client-simulator.js
```diff
@@ -0,0 +1,256 @@
...
+        log.error(`Error parsing binary plist: ${e}`);
+        return;
+      }
+
+      // plistUtils.parseBinaryPlist returns an array
+      if (plist.length === 1) {
+        plist = plist[0];
+      }
+
+      for (const key of ['WIRMessageDataKey', 'WIRDestinationKey', 'WIRSocketDataKey']) {
```
mykola-mokhnach:
```
['WIRMessageDataKey', 'WIRDestinationKey', 'WIRSocketDataKey']
  .filter((x) => plist[x])
  .map((x) => {plist[x] = plist[x].toString('utf8')})
```

lib/rpc-client-simulator.js
```diff
@@ -0,0 +1,256 @@
...
+      }
+
+      // Jump forward the length of the plist
+      this.readPos += msgLength;
+
+      // Calculate how much buffer is left
+      let leftOver = this.received.length - this.readPos;
+
+      // Is there some left over?
+      if (leftOver !== 0) {
```
mykola-mokhnach: > 0

lib/rpc-client.js
```diff
@@ -39,21 +70,98 @@ export default class RpcClient {
...
+
+        this.setDataMessageHandler(msgId.toString(), errorHandler, function dataMessageHandler (value) {
+          const msg = _.truncate(_.isString(value) ? value : JSON.stringify(value), {length: 50});
+          log.debug(`Received data response from send: '${msg}'`);
+          log.debug(`Original command: ${command}`);
+          resolve(value);
+        });
+
+        // make sure the message being sent has all the information that is needed
+        if (cmd.__argument.WIRSocketDataKey.params) {
```
mykola-mokhnach: it makes sense to put `cmd.__argument.WIRSocketDataKey.params` into a separate var to simplify this block of code

lib/rpc-client.js
```diff
@@ -39,21 +70,98 @@ export default class RpcClient {
...
+        }
+        cmd.__argument.WIRSocketDataKey.id = msgId;
+        cmd.__argument.WIRSocketDataKey =
+          Buffer.from(JSON.stringify(cmd.__argument.WIRSocketDataKey));
+      }
+
+      log.debug(`Sending '${cmd.__selector}' message to remote debugger (id: ${msgId})`);
+      try {
+        const res = await this.sendMessage(cmd);
+        if (!messageHandled) {
```
mykola-mokhnach: if not? I don't quite get the condition here

lib/rpc-client.js
```diff
@@ -99,4 +207,95 @@ export default class RpcClient {
...
+        let correctAppIdKey = dict.WIRApplicationIdentifierKey;
+
+        // if this is a report of a proxy redirect from the remote debugger
+        // we want to update our dictionary and get a new app id
+        if (oldAppIdKey && correctAppIdKey !== oldAppIdKey) {
+          log.debug(`We were notified we might have connected to the wrong app. ` +
+                    `Using id ${correctAppIdKey} instead of ${oldAppIdKey}`);
+        }
+
+        applicationConnectedHandler(dict);
```
mykola-mokhnach: is this handler always sync?

index.js
```diff
@@ -1,8 +1,16 @@
...
 
 
-const { RemoteDebugger, DEBUGGER_TYPES, REMOTE_DEBUGGER_PORT } = rd;
-const { WebKitRemoteDebugger } = wrd;
+function createRemoteDebugger (opts, realDevice = false) {
+  if (realDevice) {
+    return new RemoteDebuggerRealDevice(opts);
+  } else {
+    return new RemoteDebugger(opts);
+  }
```
dpgraham: nit: I personally prefer 

```javascript
return realDevice ? 
  new RemoteDebuggerRealDevice(opts) : 
  new RemoteDebugger(opts)
```

lib/remote-debugger-message-handler.js
```diff
@@ -112,6 +112,9 @@ export default class RpcMessageHandler {
   }
 
   async dispatchDataMessage (msgId, method, params, result, error) {
+    if (!_.isUndefined(msgId)) {
```
dpgraham: I would make this `!_.isNil(msgId)`.... `isUndefined` doesn't check if it's `null`.

index.js
```diff
@@ -1,8 +1,16 @@
-import * as rd from './lib/remote-debugger';
-import * as wrd from './lib/webkit-remote-debugger';
+import { RemoteDebugger, REMOTE_DEBUGGER_PORT } from './lib/remote-debugger';
+import RemoteDebuggerRealDevice from './lib/remote-debugger-real-device';
 
 
-const { RemoteDebugger, DEBUGGER_TYPES, REMOTE_DEBUGGER_PORT } = rd;
-const { WebKitRemoteDebugger } = wrd;
+function createRemoteDebugger (opts, realDevice = false) {
```
dpgraham: I would do it like

```javascript
function createRemoteDebugger (opts, {realDevice=false}) { ... }
```

And then call it like

```javascript
createRemoteDebugger(opts, {realDevice: <true|false>})
```

lib/remote-messages.js
```diff
@@ -66,23 +65,23 @@ class RemoteMessages {
    * Action functions
    */
 
-  getFullCommand (connId, senderId, appIdKey, pageIdKey, debuggerType, method, params) {
+  getFullCommand (connId, senderId, appIdKey, pageIdKey, method, params) {
```
dpgraham: Maybe we should refactor this command to use object syntax:

```javascript
getFullCommand({connId, senderId, appIdKey, pageIdKey, method, params}) { ... }
```

lib/rpc-client-simulator.js
```diff
@@ -0,0 +1,256 @@
...
+        // Copy what's left over into a new buffer, and save it for next time
+        let chunk = Buffer.alloc(leftOver);
```
dpgraham: `const chunk`

lib/rpc-client-real-device.js
umutuzgur: That is a good idea. Will add that

lib/rpc-client.js
umutuzgur: I don't think this makes anything look nicer

lib/remote-debugger-real-device.js
umutuzgur: Makes sense. It is more explicit then

lib/rpc-client-real-device.js
umutuzgur: Good catch

lib/rpc-client-simulator.js
umutuzgur: I agree, this looks fishy. I will not touch any of this magic until we refactor it

lib/rpc-client.js
umutuzgur: Me too. Will ask Isaac when he is back. I see that it is an early return but don't quite get the logic

lib/rpc-client.js
umutuzgur: It seems to be sync

index.js
umutuzgur: Sounds good. I also like inline conditions

lib/remote-debugger-message-handler.js
umutuzgur: Good catch. Will

index.js
umutuzgur: I don't know if this makes the code look nicer. However, I have doubts about the createRemoteDebugger function is general. I looks unnecessary since it is basically a switch for constructors

lib/remote-debugger-real-device.js
umutuzgur: I agree. Will add some

lib/remote-debugger-real-device.js
umutuzgur: In `appium-xcuitest-driver`

lib/rpc-client.js
umutuzgur: Good idea

lib/rpc-client-simulator.js
umutuzgur: I will not touch there at the moment. Can play with it when we do the refactor in Simulators

umutuzgur:
@imurchie Could you go over the PR and give me your thoughts? If you think it is enough, then I can go ahead and merge it

index.js
imurchie: Yes. Otherwise it needs to be done in the driver, which exposes complexity that the driver does not need to know.

lib/rpc-client-simulator.js
imurchie: This was work in progress stuff. I'll shift it.

lib/rpc-client.js
imurchie: There are two special cases (if there is a special handler for the selector, or if it is a data message) in which what is returned from the socket is handled elsewhere (respectively, line 119 and 134 above). Otherwise the promise still needs resolving at this point (because the message is not handled).

PR #144, opened 2019-07-25
refactor: Remove useless tests

imurchie:
Get rid of the old "unit" tests, which are adequately handled by the e2e tests.


imurchie:
These would require major refactoring, to no benefit, when the shift to `appium-ios-device` is finalized.

PR #148, opened 2019-08-01
fix: Properly save reported targets

imurchie:
It turns out that on iOS 13 there is no relation between target id and page id (earlier page id `1` would be `page-1` as a target).

So handle mapping the three variables (app id, page id, target id).

Add Xcode 11/iOS 13 to Travis build.

### DIFF
--- a/lib/rpc-client.js
+++ b/lib/rpc-client.js
@@ -257,7 +257,11 @@ export default class RpcClient {
       // way to relate it back to the original message that was sent, and
       // thus to the app and page)
       this.setSpecialMessageHandler('_rpc_applicationSentData:', reject, (appId, data) => {
-        data = JSON.parse(data);
+        try {
+          data = JSON.parse(data);
+        } catch (err) {
+          return reject(err);
+        }
         const {targetInfo} = data.params;
         this.addTarget(appIdKey, pageIdKey, targetInfo);
         resolve();
--- a/lib/rpc-client.js
+++ b/lib/rpc-client.js
@@ -57,21 +57,29 @@ export default class RpcClient {
     }
   }
 
-  async waitForTarget () {
-    if (!this.shouldCheckForTarget || !this.isTargetBased) {
+  needsTarget () {
+    return this.shouldCheckForTarget && this.isTargetBased;
+  }
+
+  async waitForTarget (appIdKey, pageIdKey) {
+    if (!this.needsTarget()) {
       return;
     }
 
     await retryInterval(WAIT_FOR_TARGET_RETRIES, WAIT_FOR_TARGET_INTERVAL, () => {
-      if (_.isEmpty(this.targets)) {
+      if (_.isEmpty(this.getTarget(appIdKey, pageIdKey))) {
         throw new Error('No targets found, unable to communicate with device');
       }
     });
   }
 
   async send (command, opts = {}) {
+    const {
+      appIdKey,
+      pageIdKey
+    } = opts;
     try {
-      await this.waitForTarget();
+      await this.waitForTarget(appIdKey, pageIdKey);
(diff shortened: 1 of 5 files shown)


lib/rpc-client.js
```diff
@@ -238,6 +249,38 @@ export default class RpcClient {
...
+        data = JSON.parse(data);
+        const {targetInfo} = data.params;
+        this.addTarget(appIdKey, pageIdKey, targetInfo);
+        resolve();
+      });
+      await this.send('setSenderKey', {
+        appIdKey,
+        pageIdKey,
+      });
+      log.debug('Sender key set');
```
mykola-mokhnach: should we print the key and page id itself

lib/rpc-client.js
```diff
@@ -238,6 +249,38 @@ export default class RpcClient {
...
+      }
+    }).finally(() => {
+      this.deleteSpecialMessageHandler('_rpc_applicationSentData:');
+    });
+    this.shouldCheckForTarget = true;
+    await this.send('enablePage', {
+      appIdKey,
+      pageIdKey,
+    });
+    log.debug('Enabled activity on page');
```
mykola-mokhnach: should we print the page id?

test/functional/safari-e2e-specs.js
```diff
@@ -76,7 +77,9 @@ describe('Safari remote debugger', function () {
     await openUrl(sim.udid, address);
   });
   afterEach(async function () {
-    await rd.disconnect();
+    if (rd) {
+      await rd.disconnect();
```
mykola-mokhnach: it probably makes sense to assign rd back to null after disconnect

lib/rpc-client.js
```diff
@@ -238,6 +249,38 @@ export default class RpcClient {
...
 
+  async selectPage (appIdKey, pageIdKey) {
+    await new B(async (resolve, reject) => {
+      // this needs to be specially set here, rather than using the usual
+      // machinery, because the response is a generic data response, and it
+      // does not include the corresponding message id (i.e., there is no
+      // way to relate it back to the original message that was sent, and
+      // thus to the app and page)
+      this.setSpecialMessageHandler('_rpc_applicationSentData:', reject, (appId, data) => {
+        data = JSON.parse(data);
```
mykola-mokhnach: can we expect an exception here?

lib/rpc-client.js
imurchie: They are printed before calling this function, and if a target is received that info is printed when it is processed.
```
dbug RemoteDebugger Selecting page '1' on app 'PID:38084' and forwarding socket setup
dbug RemoteDebugger Sending '_rpc_forwardSocketSetup:' message to remote debugger (id: 2)
dbug RemoteDebugger Sender key set
dbug RemoteDebugger Target created for app 'PID:38084' and page '1': {"targetId":"page-1","type":"page"}
```

lib/rpc-client.js
imurchie: See above.

lib/rpc-client.js
imurchie: `¯\_(ツ)_/¯`

It's Apple, so there's always the chance they messed up their own protocol.

lib/rpc-client.js
```diff
@@ -257,7 +257,11 @@ export default class RpcClient {
       // way to relate it back to the original message that was sent, and
       // thus to the app and page)
       this.setSpecialMessageHandler('_rpc_applicationSentData:', reject, (appId, data) => {
-        data = JSON.parse(data);
+        try {
```
mykola-mokhnach: this was always an interesting point for me whether such function is going to be rejected automatically if an exception happens

PR #157, opened 2019-08-19
fix: Wait between retries

umutuzgur:


lib/remote-debugger.js
```diff
@@ -303,18 +303,18 @@ class RemoteDebugger extends events.EventEmitter {
...
 
-            // we have gotten the correct application by this point, so short circuit everything
-            break appLoop;
+            if (currentUrl) {
+              log.debug(`Received app, but expected url ('${currentUrl}') was not found. Trying again.`);
+            } else {
+              log.debug('Received app, but no match was found. Trying again.');
+            }
+            pageDict = null;
+            await B.delay(500);
```
mykola-mokhnach: should we only wait here or in the `catch` block as well?

lib/remote-debugger.js
```diff
@@ -303,18 +303,18 @@ class RemoteDebugger extends events.EventEmitter {
...
-              if (currentUrl) {
-                log.debug(`Received app, but expected url ('${currentUrl}') was not found. Trying again.`);
-              } else {
-                log.debug('Received app, but no match was found. Trying again.');
-              }
-              pageDict = null;
-              continue;
+            if (found) {
+              // we have gotten the correct application by this point, so short circuit everything
+              break appLoop;
```
mykola-mokhnach: these nested loops are too big and hard to read. How about breaking them into smaller method calls, so the stuff is easier to read?

umutuzgur: I agree. I will send a cleanup PR afterwards

PR #158, opened 2019-08-20
refactor: Simplify the search app and page logics

umutuzgur:


lib/remote-debugger.js
```diff
@@ -360,6 +305,61 @@ class RemoteDebugger extends events.EventEmitter {
     }
   }
 
+  async searchApp (currentUrl, maxTries, ignoreAboutBlankUrl) {
+    for (let i = 0; i < maxTries; i++) {
```
mykola-mokhnach: how about using retryInterval from asyncbox?

lib/remote-debugger.js
```diff
@@ -360,6 +305,61 @@ class RemoteDebugger extends events.EventEmitter {
...
+      log.debug(`Trying out the possible app ids: ${possibleAppIds.join(', ')}`);
+      for (const attemptedAppIdKey of possibleAppIds) {
+        try {
+          log.debug(`Selecting app ${attemptedAppIdKey} (try #${i + 1} of ${maxTries})`);
+          const [appIdKey, pageDict] = await this.rpcClient.selectApp(attemptedAppIdKey, this.onAppConnect.bind(this));
+          // in iOS 8.2 the connect logic happens, but with an empty dictionary
+          // which leads to the remote debugger getting disconnected, and into a loop
+          if (_.isEmpty(pageDict)) {
+            log.debug('Empty page dictionary received. Trying again.');
+            continue;
```
mykola-mokhnach: should we also wait here?

lib/remote-debugger.js
umutuzgur: I'm thinking about that. I need to introduce some error handling thou. Will do that afterwards

lib/remote-debugger.js
umutuzgur: We can but it can wait for retryInterval implementation

imurchie:
💯

PR #159, opened 2019-08-21
refactor: Use retryInterval instead of for loop

umutuzgur:


lib/remote-debugger.js
```diff
@@ -306,42 +308,46 @@ class RemoteDebugger extends events.EventEmitter {
...
           }
-        } catch (err) {
-          log.debug(`Error checking application: '${err.message}'. Retrying connection`);
         }
-      }
-      await B.delay(500);
+        retryCount++;
+        throw new Error('Failed to find an app to select');
+      }, 0);
+    } catch (ign) {
```
mykola-mokhnach: I'd rather log the actual error.stack. This might be useful to debug a possible unexpected error in wrapped methods

PR #160, opened 2019-09-03
fix: compute target if we can

imurchie:
On iOS 12 the targets are always `page-<page id>`, while the id can be different, and needs to be waited for.

So, when we can, just compute the target id.

### DIFF
--- a/lib/rpc-client.js
+++ b/lib/rpc-client.js
@@ -66,6 +66,14 @@ export default class RpcClient {
       return;
     }
 
+    // on iOS less than 13 have targets that can be computed easily
+    // and sometimes is not reported by the Web Inspector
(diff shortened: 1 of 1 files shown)


lib/rpc-client.js
```diff
@@ -66,6 +66,14 @@ export default class RpcClient {
       return;
     }
 
+    // on iOS less than 13 have targets that can be computed easily
+    // and sometimes is not reported by the Web Inspector
+    if (parseInt(this.platformVersion, 10) < 13 && _.isEmpty(this.getTarget(appIdKey, pageIdKey))) {
```
mykola-mokhnach: I'd rather use util.compareVersions

PR #164, opened 2019-09-05
fix: use promises instead of http/https for async execute

imurchie:
iOS 13 stopped allowing us to install a self-signed certificate, at least in the way we had been. This means that asynchronous execution also stopped working (see https://github.com/appium/appium/issues/13184). 

@umutuzgur found the `Runtime.awaitPromise` function in [WebKit](https://github.com/WebKit/webkit/blob/master/Source/JavaScriptCore/inspector/protocol/Runtime.json#L230-L245) and this PR implements the functionality of asynchronous execution using it.

Unfortunately the function is relatively new, so versions that don't have it fall back to polling for the result to exist. Luckily timeouts are still handled by the Selenium atom.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -4,10 +4,13 @@ import _ from 'lodash';
 import assert from 'assert';
 import B from 'bluebird';
 import { util } from 'appium-support';
+import { errorFromMJSONWPStatusCode } from 'appium-base-driver';
 
 
 const WEB_CONTENT_BUNDLE_ID = 'com.apple.WebKit.WebContent';
 const MIN_PLATFORM_FOR_TARGET_BASED = '12.2';
+
+const RESPONSE_LOG_LENGTH = 100;
 
 /*
  * Takes a dictionary from the remote debugger and makes a more manageable
@@ -221,8 +224,33 @@ function getElapsedTime (startTime) {
   return parseInt((endTime[0] * 1e9 + endTime[1]) / 1e6, 10);
 }
 
+function convertResult (res) {
+  if (_.isUndefined(res)) {
+    throw new Error(`Did not get OK result from remote debugger. Result was: ${_.truncate(simpleStringify(res), {length: RESPONSE_LOG_LENGTH})}`);
+  } else if (_.isString(res)) {
+    try {
+      res = JSON.parse(res);
+    } catch (err) {
+      // we might get a serialized object, but we might not
+      // if we get here, it is just a value
+    }
+  } else if (!_.isObject(res)) {
+    throw new Error(`Result has unexpected type: (${typeof res}).`);
+  }
+
+  if (res.status && res.status !== 0) {
+    // we got some form of error.
+    throw errorFromMJSONWPStatusCode(res.status, res.value.message || res.value);
+  }
+
+  // with either have an object with a `value` property (even if `null`),
+  // or a plain object
+  return _.has(res, 'value') ? res.value : res;
+}
+
 export {
   appInfoFromDict, pageArrayFromDict, getDebuggerAppKey,
   getPossibleDebuggerAppKeys, checkParams, wrapScriptForFrame, getScriptForAtom,
   simpleStringify, deferredPromise, isTargetBased, getElapsedTime,
+  convertResult, RESPONSE_LOG_LENGTH,
 };
--- a/lib/remote-debugger-message-handler.js
+++ b/lib/remote-debugger-message-handler.js
@@ -181,27 +181,6 @@ export default class RpcMessageHandler {
     let msgId = (dataKey.id || '').toString();
     let result = dataKey.result;
 
-    // we can get an error, or we can get a response that is an error
-    let error = dataKey.error || null;
-    if (result && result.wasThrown) {
-      let message = (result.result && (result.result.value || result.result.description))
-        ? (result.result.value || result.result.description)
-        : 'Error occurred in handling data message';
-      error = new Error(message);
-    }
-
-    if (error) {
-      if (this.hasErrorHandler(msgId)) {
-        this.errorHandlers[msgId](error);
-      } else {
-        log.error(`Error occurred in handling data message: ${error}`);
-        log.error('No error handler present, ignoring');
-      }
-
-      // short circuit
-      return;
-    }
-
     let method = dataKey.method;
     let params;
     if (this.isTargetBased) {
@@ -240,6 +219,27 @@ export default class RpcMessageHandler {
       }
     } else {
       params = dataKey.params;
+    }
+
+    // we can get an error, or we can get a response that is an error
+    let error = dataKey.error || null;
+    if (result && result.wasThrown) {
+      let message = (result.result && (result.result.value || result.result.description))
+        ? (result.result.value || result.result.description)
+        : 'Error occurred in handling data message';
+      error = new Error(message);
+    }
+
+    if (error) {
+      if (this.hasErrorHandler(msgId)) {
+        this.errorHandlers[msgId](error);
+      } else {
+        log.error(`Error occurred in handling data message: ${error}`);
+        log.error('No error handler present, ignoring');
+      }
+
+      // short circuit
+      return;
     }
 
     if (!_.isEmpty(msgId)) {
--- a/lib/remote-debugger-message-handler.js
+++ b/lib/remote-debugger-message-handler.js
@@ -224,29 +224,27 @@ export default class RpcMessageHandler {
     // we can get an error, or we can get a response that is an error
     let error = dataKey.error || null;
     if (result && result.wasThrown) {
-      let message = (result.result && (result.result.value || result.result.description))
+      const message = (result.result && (result.result.value || result.result.description))
         ? (result.result.value || result.result.description)
         : 'Error occurred in handling data message';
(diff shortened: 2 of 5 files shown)


imurchie:
`appium-xcuitest-driver` web functionality remains working: https://travis-ci.org/appium/appium-xcuitest-driver/builds/581398262

lib/remote-debugger.js
```diff
@@ -403,13 +403,63 @@ class RemoteDebugger extends events.EventEmitter {
...
-    let asyncCallBack = `function (res) { xmlHttp = new XMLHttpRequest(); ` +
-                        `xmlHttp.open('POST', '${responseUrl}', true);` +
-                        `xmlHttp.setRequestHeader('Content-type','application/json'); ` +
-                        `xmlHttp.send(res); }`;
-    let script = await getScriptForAtom(atom, args, frames, asyncCallBack);
-    await this.execute(script);
+  async executeAtomAsync (atom, args, frames) {
+    // first create a Promise on the page, saving the resolve/reject functions
+    // as properties
+    const promiseName = `appiumAsyncExecutePromise${Date.now()}`;
```
mykola-mokhnach: is Date.now unique enough to guarantee there will be no duplicates? I'd rather use uuid

lib/remote-debugger.js
```diff
@@ -403,13 +403,63 @@ class RemoteDebugger extends events.EventEmitter {
...
+        appIdKey: this.appIdKey,
+        pageIdKey: this.pageIdKey,
+      });
+    } catch (err) {
+      if (!err.message.includes(`'Runtime.awaitPromise' was not found`)) {
+        throw err;
+      }
+
+      // awaitPromise is not always available, so simulate it!
+      while (true) { // eslint-disable-line no-constant-condition
```
mykola-mokhnach: should this loop have a timeout to avoid it looping forever?
should there also be a delay between loops?

lib/remote-debugger.js
```diff
@@ -403,13 +403,63 @@ class RemoteDebugger extends events.EventEmitter {
...
+        // including a timeout error, so just go until it finishes
+        if (await this.executeAtom('execute_script', [`return window.hasOwnProperty('${promiseName}Value');`, []], frames)) {
+          // we only put the property on `window` when the callback is called,
+          // so if it is there, everything is done
+          res = await this.executeAtom('execute_script', [`return window.${promiseName}Value;`, []], frames);
+          break;
+        }
+      }
+    }
+    res = JSON.parse(res);
```
mykola-mokhnach: what if res cannot be parsed?

test/functional/safari-e2e-specs.js
```diff
@@ -36,7 +36,7 @@ async function deleteDeviceWithRetry (udid) {
 
 describe('Safari remote debugger', function () {
   this.timeout(480000);
-  this.retries(2);
+  // this.retries(2);
```
mykola-mokhnach: is the comment needed?

lib/remote-debugger.js
```diff
@@ -403,13 +403,63 @@ class RemoteDebugger extends events.EventEmitter {
...
+      }
+
+      // awaitPromise is not always available, so simulate it!
+      while (true) { // eslint-disable-line no-constant-condition
+        // the atom _will_ return, either because it finished or an error
+        // including a timeout error, so just go until it finishes
+        if (await this.executeAtom('execute_script', [`return window.hasOwnProperty('${promiseName}Value');`, []], frames)) {
+          // we only put the property on `window` when the callback is called,
+          // so if it is there, everything is done
+          res = await this.executeAtom('execute_script', [`return window.${promiseName}Value;`, []], frames);
```
mykola-mokhnach: should we also detach the unused promise from the window after the result is received?

lib/remote-debugger-message-handler.js
```diff
@@ -242,6 +221,27 @@ export default class RpcMessageHandler {
...
+    if (error) {
+      if (this.hasErrorHandler(msgId)) {
+        this.errorHandlers[msgId](error);
+      } else {
+        log.error(`Error occurred in handling data message: ${error}`);
+        log.error('No error handler present, ignoring');
+      }
+
+      // short circuit
+      return;
```
umutuzgur: No longer needed

lib/remote-debugger.js
```diff
@@ -403,13 +403,63 @@ class RemoteDebugger extends events.EventEmitter {
...
+    // as properties
+    const promiseName = `appiumAsyncExecutePromise${Date.now()}`;
+    let script = `var res, rej;` +
+                 `window.${promiseName} = new Promise(function (resolve, reject) {` +
+                 `  res = resolve;` +
+                 `  rej = reject;` +
+                 `});` +
+                 `window.${promiseName}.resolve = res;` +
+                 `window.${promiseName}.reject = rej;` +
+                 `window.${promiseName};`;
```
mykola-mokhnach: what this step is for?

lib/remote-debugger.js
umutuzgur: You cant really. It is a JS object and we van have `-` in the object name

lib/remote-debugger.js
umutuzgur: I agree. Let's use retryInterval

lib/remote-debugger.js
mykola-mokhnach: You can take just first 8 chars - my intention here is to make it unique. There is still a theoretical possibility this method is called twice with different params at the same millisecond, so then the second call will override the first one if date stamp is used

lib/remote-debugger.js
umutuzgur: I guess to return the promise as the last line in a method doesn't need return in JS. However, it would be nice to have a `return` to make it explicit

lib/remote-debugger.js
umutuzgur: Can't we use `convertResult` function here?

lib/remote-debugger-message-handler.js
```diff
@@ -242,6 +221,27 @@ export default class RpcMessageHandler {
       params = dataKey.params;
     }
 
+    // we can get an error, or we can get a response that is an error
+    let error = dataKey.error || null;
+    if (result && result.wasThrown) {
```
umutuzgur: I would change this method now that if there is no error, we return early

lib/remote-debugger.js
imurchie: It is not run in a function, so a return will error, unfortunately. I’ll add a comment to this effect.

lib/remote-debugger.js
imurchie: Sure. Though the chances of uuid clash and this being called twice in the same ms are probably of the same order.

mykola-mokhnach: nah, cannot be (of course if uuid generation is implemented properly)

imurchie: Nor can this function be called twice in the same context at the same ms.

lib/remote-messages.js
```diff
@@ -83,7 +83,7 @@ class RemoteMessages {
     const method = 'Runtime.evaluate';
     const params = {
       expression: opts.command,
-      returnByValue: true,
+      returnByValue: _.isBoolean(opts.returnByValue) ? opts.returnByValue : true,
```
umutuzgur: We can also just do a `!!opts.returnByValue`

imurchie: Except it should default to `true`.

lib/helpers.js
```diff
@@ -221,8 +224,33 @@ function getElapsedTime (startTime) {
   return parseInt((endTime[0] * 1e9 + endTime[1]) / 1e6, 10);
 }
 
+function convertResult (res) {
+  if (_.isUndefined(res)) {
+    throw new Error(`Did not get OK result from remote debugger. Result was: ${_.truncate(simpleStringify(res), {length: RESPONSE_LOG_LENGTH})}`);
```
mykola-mokhnach: emmm... The `res` is undefined here, what do we want to log though?

lib/helpers.js
```diff
@@ -221,8 +224,33 @@ function getElapsedTime (startTime) {
...
+  } else if (_.isString(res)) {
+    try {
+      res = JSON.parse(res);
```
mykola-mokhnach: there is safeJsonParse function in util

lib/helpers.js
```diff
@@ -221,8 +224,33 @@ function getElapsedTime (startTime) {
...
+    } catch (err) {
+      // we might get a serialized object, but we might not
+      // if we get here, it is just a value
+    }
+  } else if (!_.isObject(res)) {
```
mykola-mokhnach: perhaps isPlainObject?

lib/helpers.js
```diff
@@ -221,8 +224,33 @@ function getElapsedTime (startTime) {
...
+    throw new Error(`Result has unexpected type: (${typeof res}).`);
+  }
+
+  if (res.status && res.status !== 0) {
```
mykola-mokhnach: `res.status` will be already false if it is equal to zero

mykola-mokhnach: `res.status` will fail if res is null

PR #165, opened 2019-09-25
fix: connection issues

imurchie:
**Breaking change**

Simplify how connected apps are treated. Mostly, do not wait after setting the connection key in the Web Inspector, then allow them to be handled asynchronously.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -21,6 +21,13 @@ function appInfoFromDict (dict) {
   const isProxy = _.isString(dict.WIRIsApplicationProxyKey)
     ? dict.WIRIsApplicationProxyKey.toLowerCase() === 'true'
     : dict.WIRIsApplicationProxyKey;
+  const isAutomationEnabled = _.has(dict, 'WIRAutomationAvailabilityKey')
+    ? _.isString(dict.WIRAutomationAvailabilityKey)
+      ? dict.WIRAutomationAvailabilityKey === 'WIRAutomationAvailabilityUnknown'
+        ? 'Unknown'
+        : dict.WIRAutomationAvailabilityKey === 'WIRAutomationAvailabilityAvailable'
+      : dict.WIRAutomationAvailabilityKey
+    : !!dict.WIRRemoteAutomationEnabledKey;
   const entry = {
     id,
     isProxy,
@@ -28,7 +35,7 @@ function appInfoFromDict (dict) {
     bundleId: dict.WIRApplicationBundleIdentifierKey,
     hostId: dict.WIRHostApplicationIdentifierKey,
     isActive: dict.WIRIsApplicationActiveKey,
-    isAutomationEnabled: !!dict.WIRRemoteAutomationEnabledKey,
+    isAutomationEnabled,
   };
 
   return [id, entry];
--- a/lib/rpc-client.js
+++ b/lib/rpc-client.js
@@ -73,18 +73,22 @@ export default class RpcClient {
       return;
     }
 
+    function getNewTargetId (targets) {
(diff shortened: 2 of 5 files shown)


lib/helpers.js
```diff
@@ -21,14 +21,21 @@ function appInfoFromDict (dict) {
   const isProxy = _.isString(dict.WIRIsApplicationProxyKey)
     ? dict.WIRIsApplicationProxyKey.toLowerCase() === 'true'
     : dict.WIRIsApplicationProxyKey;
+  const isAutomationEnabled = _.has(dict, 'WIRAutomationAvailabilityKey')
+    ? _.isString(dict.WIRAutomationAvailabilityKey)
```
mykola-mokhnach: Can you please add paretheses to simplify the parsing of this expression? As far as I can understand this variable can have both string or boolean value

lib/rpc-client.js
```diff
@@ -33,6 +33,7 @@ export default class RpcClient {
 
     this.connected = false;
     this.connId = UUID.create().toString();
+    // this.connId = 'F8532879-C583-45E1-974C-101AE3D8EEF3';
```
mykola-mokhnach: do we need this comment?

lib/rpc-client.js
```diff
@@ -73,18 +74,22 @@ export default class RpcClient {
       return;
     }
 
+    function getNewTargetId (targets) {
+      const targetIds = _.values(targets)
+        .map((targetId) => targetId.replace('page-', ''))
+        .map((targetId) => parseInt(targetId, 10))
```
mykola-mokhnach: should we filter out NaNs here?

lib/rpc-client.js
```diff
@@ -149,51 +157,54 @@ export default class RpcClient {
...
-          cmd.__argument.WIRSocketDataKey.params.id = msgId;
-          if (!cmd.__argument.WIRSocketDataKey.params.targetId && this.needsTarget()) {
-            cmd.__argument.WIRSocketDataKey.params.targetId = this.getTarget(sendOpts.appIdKey, sendOpts.pageIdKey);
-          }
-          if (cmd.__argument.WIRSocketDataKey.params.message) {
-            cmd.__argument.WIRSocketDataKey.params.message.id = msgId;
-            cmd.__argument.WIRSocketDataKey.params.message = JSON.stringify(cmd.__argument.WIRSocketDataKey.params.message);
+      // if we do not want to wait for response, just pass through here
+      if (waitForResponse) {
+        if (this.messageHandler.hasSpecialMessageHandler(cmd.__selector)) {
```
mykola-mokhnach: can this block be extracted to a separate function?

lib/rpc-client.js
```diff
@@ -136,7 +144,7 @@ export default class RpcClient {
         : _.truncate(JSON.stringify(res), DATA_LOG_LENGTH);
   }
 
-  async sendToDevice (command, opts = {}) {
+  async sendToDevice (command, opts = {}, waitForResponse = true) {
```
mykola-mokhnach: should `waitForResponse` be an option?

lib/rpc-client.js
imurchie: No. These are logged as they come in, and if they are not numbers then something has changed in the protocol and the error will be useful. As there is no indication that that will happen, I don't feel the need to program for it now.

lib/rpc-client.js
imurchie: The options are sent to the Web Inspector. `waitForResponse` is used locally to adjust the sending itself.

imurchie:
Whoops! Merged too soon. I'll open another PR.

PR #166, opened 2019-09-26
refactor: clean up send and parse

imurchie:
Mostly address comments from #165 which I merged accidentally.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -21,13 +21,19 @@ function appInfoFromDict (dict) {
   const isProxy = _.isString(dict.WIRIsApplicationProxyKey)
     ? dict.WIRIsApplicationProxyKey.toLowerCase() === 'true'
     : dict.WIRIsApplicationProxyKey;
-  const isAutomationEnabled = _.has(dict, 'WIRAutomationAvailabilityKey')
-    ? _.isString(dict.WIRAutomationAvailabilityKey)
-      ? dict.WIRAutomationAvailabilityKey === 'WIRAutomationAvailabilityUnknown'
+  // automation enabled can be either from the keys
+  //   - WIRRemoteAutomationEnabledKey (boolean)
+  //   - WIRAutomationAvailabilityKey (string or boolean)
+  let isAutomationEnabled = !!dict.WIRRemoteAutomationEnabledKey;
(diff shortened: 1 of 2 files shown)


lib/helpers.js
```diff
@@ -21,13 +21,19 @@ function appInfoFromDict (dict) {
   const isProxy = _.isString(dict.WIRIsApplicationProxyKey)
     ? dict.WIRIsApplicationProxyKey.toLowerCase() === 'true'
     : dict.WIRIsApplicationProxyKey;
-  const isAutomationEnabled = _.has(dict, 'WIRAutomationAvailabilityKey')
-    ? _.isString(dict.WIRAutomationAvailabilityKey)
-      ? dict.WIRAutomationAvailabilityKey === 'WIRAutomationAvailabilityUnknown'
+  // automation enabled can be either from the keys
+  //   - WIRRemoteAutomationEnabledKey (boolean)
+  //   - WIRAutomationAvailabilityKey (string or boolean)
+  let isAutomationEnabled = !!dict.WIRRemoteAutomationEnabledKey;
```
mykola-mokhnach: nit: can we then have a better variable name? `is` means it is strictly boolean

imurchie: The only case where it is not boolean is when it is unknown. Which is a temporary state while the Web Inspector gets its act in gear and reports properly.

lib/rpc-client.js
```diff
@@ -156,55 +155,54 @@ export default class RpcClient {
...
-            if (!cmd.__argument.WIRSocketDataKey.params.targetId && this.needsTarget()) {
-              cmd.__argument.WIRSocketDataKey.params.targetId = this.getTarget(sendOpts.appIdKey, sendOpts.pageIdKey);
-            }
-            if (cmd.__argument.WIRSocketDataKey.params.message) {
-              cmd.__argument.WIRSocketDataKey.params.message.id = msgId;
-              cmd.__argument.WIRSocketDataKey.params.message = JSON.stringify(cmd.__argument.WIRSocketDataKey.params.message);
-            }
+      let messageHandled = true;
+      if (!waitForResponse) {
+        // the promise will be resolved as soon as the socket has been sent
```
mykola-mokhnach: hm, "socket sent" ?

PR #172, opened 2019-10-28
fix: make sure that app dict is not modified at the same time

imurchie:
Use `async-lock` to make sure that the callbacks that adjust the application dictionary do not do so at the same time.

Also do a couple of other things:
1. Change `logFullResponse ` to `logAllCommunication` and use `appium-ios-device` to do this logging. This will continue to work through the calling code.
1. Ensure `appium-ios-device` gets the `isSimulator` parameter for creating Web Inspector service.

This ought to help support lower iOS versions (see https://github.com/appium/appium/issues/13388). Tests for iOS 9.3 fail with current master (https://travis-ci.org/appium/appium-remote-debugger/jobs/603818016) but work with this branch (https://travis-ci.org/appium/appium-remote-debugger/jobs/603816787).

### DIFF
--- a/bin/web_inspector_proxy.js
+++ b/bin/web_inspector_proxy.js
@@ -1,23 +1,11 @@
 /* eslint-disable no-console */
 import { exec, SubProcess } from 'teen_process';
-import { plist } from 'appium-support';
+import { plist, util } from 'appium-support';
 import { asyncify } from 'asyncbox';
 import B from 'bluebird';
 import { getSimulator } from 'appium-ios-simulator';
 import _ from 'lodash';
 
-
-function logFullMessage (plist) {
-  // Buffers cannot be serialized in a readable way
-  const bufferToJSON = Buffer.prototype.toJSON;
-  delete Buffer.prototype.toJSON;
-  try {
-    console.log(JSON.stringify(plist, (k, v) => Buffer.isBuffer(v) ? v.toString('utf8') : v, 2));
-  } finally {
-    // restore the function, so as to not break further serialization
-    Buffer.prototype.toJSON = bufferToJSON;
-  }
-}
 
 async function getSocket (udid) {
   const sim = await getSimulator(udid);
@@ -61,7 +49,7 @@ function printRecord (lines) {
     const buf = Buffer.from(data);
     try {
       const doc = plist.parsePlist(buf);
-      logFullMessage(doc);
+      console.log(util.jsonStringify(doc));
     } catch (err) {
       if (err.message.includes('maxObjectCount exceeded')) {
         return str;
--- a/lib/message-handlers.js
+++ b/lib/message-handlers.js
@@ -17,23 +17,26 @@ async function onPageChange (appIdKey, pageDict) {
   const pageArray = pageArrayFromDict(pageDict);
 
   await this.useAppDictLock((done) => {
-    // save the page dict for this app
-    if (this.appDict[appIdKey]) {
-      if (this.appDict[appIdKey].pageArray) {
-        if (this.appDict[appIdKey].pageArray.resolve) {
-          // pageDict is a pending promise, so resolve
-          this.appDict[appIdKey].pageArray.resolve();
-        } else {
-          // we have a pre-existing pageDict
-          if (_.isEqual(this.appDict[appIdKey].pageArray, pageArray)) {
-            log.debug(`Received page change notice for app '${appIdKey}' ` +
-                      `but the listing has not changed. Ignoring.`);
-            return done();
+    try {
+      // save the page dict for this app
+      if (this.appDict[appIdKey]) {
+        if (this.appDict[appIdKey].pageArray) {
+          if (this.appDict[appIdKey].pageArray.resolve) {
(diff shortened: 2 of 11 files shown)


lib/remote-debugger.js
```diff
@@ -741,6 +747,10 @@ class RemoteDebugger extends events.EventEmitter {
       }
     });
   }
+
+  useAppDictLock (fn) {
+    this._lock.acquire('appDict', fn);
```
mykola-mokhnach: shouldn't this be a promise?

mykola-mokhnach:
```
async useAppDictLock (fn) {
     return await this._lock.acquire('appDict', fn);
}
```

bin/web_inspector_proxy.js
```diff
@@ -61,7 +49,7 @@ function printRecord (lines) {
     const buf = Buffer.from(data);
     try {
       const doc = plist.parsePlist(buf);
-      logFullMessage(doc);
+      console.log(util.jsonStringify(doc));
```
mykola-mokhnach: should console.log be here?

lib/message-handlers.js
```diff
@@ -9,32 +9,34 @@ import _ from 'lodash';
...
-      if (this.appDict[appIdKey].pageArray.resolve) {
-        // pageDict is a pending promise, so resolve
-        this.appDict[appIdKey].pageArray.resolve();
-      } else {
-        // we have a pre-existing pageDict
-        if (_.isEqual(this.appDict[appIdKey].pageArray, pageArray)) {
-          log.debug(`Received page change notice for app '${appIdKey}' ` +
-                    `but the listing has not changed. Ignoring.`);
-          return;
+  this.useAppDictLock((done) => {
```
mykola-mokhnach: shouldn't it be awaited?

lib/remote-debugger.js
```diff
@@ -741,6 +747,10 @@ class RemoteDebugger extends events.EventEmitter {
       }
     });
   }
+
+  useAppDictLock (fn) {
```
mykola-mokhnach: this function could be a simple promise, so then we should not play all this stuff about `done` callback

lib/remote-debugger.js
imurchie: Indeed it should.

bin/web_inspector_proxy.js
imurchie: Yes, this is a command line tool.

lib/message-handlers.js
```diff
@@ -49,11 +51,13 @@ function onPageChange (appIdKey, pageDict) {
...
 
-function onAppConnect (dict) {
+async function onAppConnect (dict) {
   let appIdKey = dict.WIRApplicationIdentifierKey;
   log.debug(`Notified that new application '${appIdKey}' has connected`);
-
-  this.updateAppsWithDict(dict);
+  await this.useAppDictLock((done) => {
+    this.updateAppsWithDict(dict);
+    done();
```
mykola-mokhnach: should we call done in finally block ?

lib/message-handlers.js
```diff
@@ -110,7 +117,10 @@ function onConnectedApplicationList (apps) {
     newDict[id] = entry;
   }
   // update the object's list of apps
-  _.defaults(this.appDict, newDict);
+  await this.useAppDictLock((done) => {
```
mykola-mokhnach: `await this.useAppDictLock(async () => await _.defaults(this.appDict, newDict))` should in theory work as well

mykola-mokhnach: See `promise mode` at https://www.npmjs.com/package/async-lock

imurchie: It seems weird to me to await a non-async function

lib/remote-debugger.js
imurchie: It would be adding promise-handling for the sake of removing a callback. Seems like a step backwards

lib/message-handlers.js
```diff
@@ -9,32 +9,37 @@ import _ from 'lodash';
...
-      if (this.appDict[appIdKey].pageArray.resolve) {
-        // pageDict is a pending promise, so resolve
-        this.appDict[appIdKey].pageArray.resolve();
-      } else {
-        // we have a pre-existing pageDict
-        if (_.isEqual(this.appDict[appIdKey].pageArray, pageArray)) {
-          log.debug(`Received page change notice for app '${appIdKey}' ` +
-                    `but the listing has not changed. Ignoring.`);
-          return;
+  await this.useAppDictLock((done) => {
```
jlipps: this is readable enough as-is, but i wonder if it would make sense to revisit our lock code in general, to see if we could have a model like this?

```js
await this.appDictLock.acquire(); // wait to use the app dict until we know it's not locked
try {
  // do things now that we have the lock
} finally {
  this.appDictLock.release(); // say we don't need the resource anymore
}
```

imurchie: I would be fine with that style, though I think it should be wrapped and put into `appium-support` if it were to be implemented.

PR #174, opened 2019-11-11
Exclude the treatment of WebViews from non-active applications

anorimaki:
Try to fix https://github.com/appium/appium/issues/13569


lib/remote-debugger.js
```diff
@@ -267,7 +267,7 @@ class RemoteDebugger extends events.EventEmitter {
 
       let fullPageArray = [];
       for (const [app, info] of _.toPairs(this.appDict)) {
-        if (!_.isArray(info.pageArray)) {
+        if (!_.isArray(info.pageArray) || (info.isActive === 0)) {
```
jlipps: parens around the latter part are not necessary

PR #175, opened 2019-11-12
fix: use web inspector service verbose and chunk size parameters

imurchie:
Pass through the logging of communication, and chunk size.

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -1,5 +1,15 @@
 import { logger } from 'appium-support';
+import moment from 'moment';
 
-const log = logger.getLogger('RemoteDebugger');
+
+const FORCE_LOG_MODE = parseInt(process.env._FORCE_LOGS, 10) === 1;
+
+const log = logger.getLogger(function () {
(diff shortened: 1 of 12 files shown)


lib/logger.js
```diff
@@ -1,5 +1,15 @@
 import { logger } from 'appium-support';
+import moment from 'moment';
```
mykola-mokhnach: is the package listed in package.json?

lib/logger.js
```diff
@@ -1,5 +1,15 @@
...
 
-const log = logger.getLogger('RemoteDebugger');
+
+const FORCE_LOG_MODE = parseInt(process.env._FORCE_LOGS, 10) === 1;
+
+const log = logger.getLogger(function () {
+  if (FORCE_LOG_MODE) {
```
mykola-mokhnach: how about using the ternary operator here?

imurchie: This stuff doesn't need to be here anyway. Deleting.

PR #176, opened 2019-11-12
fix: enable console logging on page after navigating

imurchie:
Without re-enabling console logging, the events only come if we are connected to a local server.

Fixes https://github.com/appium/appium/issues/13431


lib/remote-debugger.js
```diff
@@ -172,7 +172,9 @@ class RemoteDebugger extends events.EventEmitter {
   }
 
   async disconnect () {
-    await this.rpcClient.disconnect();
+    if (this.rpcClient) {
+      await this.rpcClient.disconnect();
```
mykola-mokhnach: should we also set this.rpcClient to null?

imurchie: That happens in the `teardown` function, called later in this function.

PR #177, opened 2019-11-15
fix: do not use user gestures

imurchie:
I had set the `emulateUserGesture` flag to `true` since it is what Safari sends. But in iOS 13+ this somehow bypasses the popup blocking in MobileSafari (so even set to true, new windows can be open).

I had also accidentally removed the prefix to the logger.

### DIFF
--- a/lib/remote-messages.js
+++ b/lib/remote-messages.js
@@ -76,6 +76,14 @@ class RemoteMessages {
   getFullCommand (opts = {}) {
     const {method, params, connId, senderId, appIdKey, pageIdKey, targetId, id} = opts;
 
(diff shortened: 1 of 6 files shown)


lib/remote-messages.js
```diff
@@ -90,7 +90,7 @@ class RemoteMessages {
             objectGroup: 'console',
             includeCommandLineAPI: true,
             doNotPauseOnExceptionsAndMuteConsole: false,
-            emulateUserGesture: true,
+            emulateUserGesture: false,
```
mykola-mokhnach: it would be nice to have some explanation comment here

imurchie: Done.

PR #178, opened 2019-11-20
refactor: move callbacks into correct places

imurchie:
There were a number of places where callbacks were in the wrong place, so, for instance, an event would come in and be passed up to the RemoteDebugger from the RPC client, only to be passed back down to the RPC client.

So, move some of those around into the right places.

### DIFF
--- a/lib/remote-debugger.js
+++ b/lib/remote-debugger.js
@@ -107,6 +107,7 @@ class RemoteDebugger extends events.EventEmitter {
     this.pageIdKey = null;
     this.pageLoading = false;
     this._navigatingToPage = false;
+    this.allowNavigationWithoutReload = false;
 
     // set up the special callbacks for handling rd events
     this.specialCbs = {
@@ -119,8 +120,6 @@ class RemoteDebugger extends events.EventEmitter {
       '_rpc_reportConnectedDriverList:': this.onConnectedDriverList.bind(this),
       pageLoad: this.pageLoad.bind(this),
       frameDetached: this.frameDetached.bind(this),
-      targetCreated: this.onTargetCreated.bind(this),
-      targetDestroyed: this.onTargetDestroyed.bind(this),
     };
 
     this.rpcClient = null;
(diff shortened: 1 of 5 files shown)


lib/remote-debugger.js
```diff
@@ -596,11 +601,17 @@ class RemoteDebugger extends events.EventEmitter {
       // add a handler for the `Page.frameNavigated` message
       // from the remote debugger
       const navEventListener = (value) => {
-        log.debug(`Frame navigated in ${((Date.now() - startMs) / 1000)} sec from source: ${value}`);
+        log.debug(`Frame navigated in ${((Date.now() - startMs) / 1000)}s from: ${value}`);
```
mykola-mokhnach: should we trunc to seconds or milliseconds here?

lib/rpc-message-handler.js
```diff
@@ -158,7 +147,7 @@ export default class RpcMessageHandler {
       this.dataHandlers[msgId](result);
       this.dataHandlers[msgId] = null;
     } else if (this.dataHandlers[msgId] === null) {
-      log.error(`Debugger returned data for message ${msgId} ` +
+      log.error(`Debugger returned data for message '${msgId}' ` +
                 `but we already ran that callback! WTF??`);
```
mykola-mokhnach: ahm, How about making the message less angry?

imurchie: I've never seen this actually happen. But it's been there since time immemorial. By which I mean March 22, 2013 (https://github.com/appium/appium/commit/0d6818d8a4a265b239e4db8717e74e0ba83b3d19). I like it.

lib/remote-debugger.js
imurchie: This is as it always has been. I'd rather not change it.

lib/rpc-message-handler.js
mykola-mokhnach: What a shame (although I like it as well :-P )

PR #179, opened 2019-11-20
feat: allow '*' as bundle id reported by Safari

imurchie:
Use apps with the bundle identifier `"*"`.

In doing this there was some code that could be yanked relating to old platforms that this version does not support.

Related to https://github.com/appium/appium/issues/13622

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -85,17 +85,17 @@ function getDebuggerAppKey (bundleId, appDict) {
   // now we need to determine if we should pick a proxy for this instead
   if (appId) {
     log.debug(`Found app id key '${appId}' for bundle '${bundleId}'`);
-    let proxiedAppIds = [];
+    let proxyAppId;
     for (const [key, data] of _.toPairs(appDict)) {
       if (data.isProxy && data.hostId === appId) {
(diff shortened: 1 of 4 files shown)


lib/helpers.js
```diff
@@ -73,35 +74,29 @@ function pageArrayFromDict (pageDict) {
...
+      if (data.isProxy && data.hostId === appId) {
+        log.debug(`Found separate bundleId '${data.bundleId}' ` +
+                  `acting as proxy for '${bundleId}', with app id '${key}'`);
+        proxiedAppIds.push(key);
       }
     }
-  } else {
-    if (_.has(appDict, bundleId)) {
-      appId = bundleId;
+    if (proxiedAppIds.length) {
```
mykola-mokhnach: we could set the app id in the previous loop. There is no need for the intermediate proxiedAppIds array

jlipps: the only thing awkward about that is the log statement has no obvious home.

imurchie: Yeah, you'd need an intermediate flag. Opted to move to just storing the proxied app id, rather than an array of them, and check that.

PR #181, opened 2019-11-28
Update proxy connection ability

christian-bromann:
Since we have implemented the ability to route the socket connection to the iOS device via a proxy some things have updated (e.g. the way we communicate to the device with `appium-ios-device`). This PR intends to update the current behavior so that a proper connection on the proxy side can be established.

__Note:__ this is work in progress but I don't think much changes will be applied at this point.

### DIFF
--- a/lib/rpc-client.js
+++ b/lib/rpc-client.js
@@ -119,15 +119,15 @@ export default class RpcClient {
     } = opts;
     try {
       await this.waitForTarget(appIdKey, pageIdKey);
-      return await this.sendToDevice(command, opts, waitForResponse);
+      return this.sendToDevice(command, opts, waitForResponse);
     } catch (err) {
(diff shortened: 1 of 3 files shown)


lib/rpc-client.js
```diff
@@ -119,15 +119,15 @@ export default class RpcClient {
     } = opts;
     try {
       await this.waitForTarget(appIdKey, pageIdKey);
-      return await this.sendToDevice(command, opts, waitForResponse);
+      return this.sendToDevice(command, opts, waitForResponse);
```
imurchie: Just as a heads up, the `return await` is the style used in Appium, so it should remain.

yashLadha: Also, it will help to catch the rejection inside this function itself.

yashLadha: Also, it will help to catch the rejection inside this function itself.

PR #183, opened 2019-12-05
Throw 'TimeoutError' on executeAtomAsync when timeout is 0ms.

mattyg:
If the script timeout is set to 0ms and Runtime.awaitPromise is not available, executeAtomAsync never actually attempts to execute the script.

This causes an unclear error to be thrown from convertResult of helpers.js:  `Result has unexpected type: (object)`. The object is null.

With this PR, if script timeout is 0ms and Runtime.awaitPromise is not available, executeAtomAsync preemptively throws a TimeoutError.


imurchie:
This has been fixed by forcing it to run at least once.

PR #189, opened 2020-01-07
fix: add safariviewcontroller bundle id to standard list

imurchie:
Automatically find SafariViewController apps if available.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -124,12 +124,10 @@ function appIdForBundle (bundleId, appDict) {
 function getPossibleDebuggerAppKeys (bundleIds, appDict) {
   let proxiedAppIds = [];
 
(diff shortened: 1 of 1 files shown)


lib/helpers.js
```diff
@@ -123,8 +124,12 @@ function appIdForBundle (bundleId, appDict) {
 function getPossibleDebuggerAppKeys (bundleIds, appDict) {
   let proxiedAppIds = [];
 
+  const extraBundleIds = bundleIds.includes(SAFARI_WEB_VIEW_BUNDLE_ID)
```
mykola-mokhnach: `extraBundleIds = _.difference([SAFARI_WEB_VIEW_BUNDLE_ID, WILDCARD_BUNDLE_ID], bundleIds)`

imurchie: Not sure what that would accomplish.

PR #192, opened 2020-01-17
Feat: add possibility of full initialization if web inspector. 

imurchie:
Much of this is refactoring to organize the messages. And then add the possibility to do every step that desktop safari does when you start a web inspector debugging session. This is turned off by default. 

Once this is in I have some work to simplify things.

### DIFF
--- a/lib/remote-messages.js
+++ b/lib/remote-messages.js
@@ -50,12 +50,16 @@ class RemoteMessages {
   }
 
   indicateWebView (opts = {}) {
-    const {connId, appIdKey, pageIdKey} = opts;
-    const {enabled} = opts.opts;
+    const {
+      connId,
+      appIdKey,
+      pageIdKey,
+      enabled,
+    } = opts;
     return {
       __argument: {
         WIRApplicationIdentifierKey: appIdKey,
-        WIRIndicateEnabledKey: _.isUndefined(enabled) ? true : enabled,
+        WIRIndicateEnabledKey: _.isNil(enabled) ? true : enabled,
         WIRConnectionIdentifierKey: connId,
         WIRPageIdentifierKey: pageIdKey
       },
@@ -108,8 +112,8 @@ class RemoteMessages {
             includeCommandLineAPI: true,
             doNotPauseOnExceptionsAndMuteConsole: false,
             emulateUserGesture: false,
-            generatePreview: true,
-            saveResult: true,
+            generatePreview: false,
+            saveResult: false,
           }
         }),
       };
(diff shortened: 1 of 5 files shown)


lib/remote-messages.js
```diff
@@ -206,100 +210,199 @@ class RemoteMessages {
     let method, params;
 
     switch (command) {
+      /* BASIC COMMANDS */
       case 'setConnectionKey':
         return this.setConnectionKey({connId});
       case 'indicateWebView':
-        return this.indicateWebView({connId, appIdKey, pageIdKey, opts});
+        return this.indicateWebView({connId, appIdKey, pageIdKey, enabled: opts.enabled});
```
mykola-mokhnach: can we unpack `enabled` above?

lib/rpc-client.js
```diff
@@ -310,26 +316,129 @@ export default class RpcClient {
...
 
+    // go through the steps that the Desktop Safari system
+    // goes through to initialize the Web Inspector session
+
     const sendOpts = {
       appIdKey,
       pageIdKey,
     };
 
+    for (const enabled of [true, false]) {
```
mykola-mokhnach: what this magic loop is doing?

lib/rpc-client.js
```diff
@@ -310,26 +316,129 @@ export default class RpcClient {
...
+      shouldPause: false,
+    }, false);
+
+    await this.send('LayerTree.enable', sendOpts, false);
+    await this.send('Worker.enable', sendOpts, false);
+    await this.send('Canvas.enable', sendOpts, false);
+    await this.send('Console.enable', sendOpts, false);
+    await this.send('DOM.getDocument', sendOpts, false);
+    const loggingChannels = await this.send('Console.getLoggingChannels', sendOpts);
+    for (const source of (loggingChannels.channels || {}).map((entry) => entry.source)) {
```
mykola-mokhnach: can `map` be applied to an object?

mykola-mokhnach:
tests, probably, need some updates

imurchie:
No need to update the tests. This is all refactoring other than adding a flag to do the I initialization.

lib/rpc-client.js
imurchie: Sending a `true` then a `false` to the same command. This highlights and then unhighlights the webview.

lib/rpc-client.js
imurchie: Nope. Fixing.

lib/rpc-client.js
mykola-mokhnach: would be nice to have some comment there

PR #194, opened 2020-01-28
refactor: use EventEmitter instead of passing callbacks all over

imurchie:
Node has a great library for [event handling](https://nodejs.org/api/events.html#events_class_eventemitter). This package has been passing callbacks around to handle all the asynchronous messaging of the Web Inspector, which has outlived its usefulness.

So replace everything with `EventEmitter`. This is a **breaking change** due to the change in external events (adding an error parameter).

After this approach is vetted and in the module, it can be extended so that users of `appium-remote-debugger` can listen for whatever event form the Web Inspector they want to implement services.

### DIFF
--- a/lib/message-handlers.js
+++ b/lib/message-handlers.js
@@ -9,7 +9,7 @@ import _ from 'lodash';
  * These will be added to the prototype.
  */
 
-async function onPageChange (appIdKey, pageDict) {
+async function onPageChange (err, appIdKey, pageDict) {
   if (_.isEmpty(pageDict)) {
     return;
   }
@@ -54,7 +54,7 @@ async function onPageChange (appIdKey, pageDict) {
   });
 }
 
-async function onAppConnect (dict) {
+async function onAppConnect (err, dict) {
   let appIdKey = dict.WIRApplicationIdentifierKey;
   log.debug(`Notified that new application '${appIdKey}' has connected`);
   await this.useAppDictLock((done) => {
@@ -66,7 +66,7 @@ async function onAppConnect (dict) {
   });
 }
 
-function onAppDisconnect (dict) {
+function onAppDisconnect (err, dict) {
   let appIdKey = dict.WIRApplicationIdentifierKey;
   log.debug(`Application '${appIdKey}' disconnected. Removing from app dictionary.`);
   log.debug(`Current app is ${this.appIdKey}`);
@@ -89,7 +89,7 @@ function onAppDisconnect (dict) {
   }
 }
 
-async function onAppUpdate (dict) {
+async function onAppUpdate (err, dict) {
   await this.useAppDictLock((done) => {
     try {
       this.updateAppsWithDict(dict);
@@ -99,7 +99,7 @@ async function onAppUpdate (dict) {
   });
 }
 
-function onConnectedDriverList (drivers) {
+function onConnectedDriverList (err, drivers) {
   this.connectedDrivers = drivers.WIRDriverDictionaryKey;
   log.debug(`Received connected driver list: ${JSON.stringify(this.connectedDrivers)}`);
 }
--- a/lib/remote-debugger.js
+++ b/lib/remote-debugger.js
@@ -182,7 +182,7 @@ class RemoteDebugger extends events.EventEmitter {
     this.teardown();
   }
 
-  isConnected () {
+  get isConnected () {
     return !!(this.rpcClient && this.rpcClient.isConnected);
   }
 
@@ -657,47 +657,41 @@ class RemoteDebugger extends events.EventEmitter {
   }
 
   /*
-   * Keep track of the console event listeners so they can be removed
+   * Keep track of the client event listeners so they can be removed
    */
-  get consoleListeners () {
-    this._consoleListeners = this._consoleListeners || [];
-    return this._consoleListeners;
+  addClientEventListener (eventName, listener) {
+    this._clientEventListeners = this._clientEventListeners || {};
+    this._clientEventListeners[eventName] = this._clientEventListeners[eventName] || [];
+    this._clientEventListeners[eventName].push(listener);
+    this.rpcClient.on(eventName, listener);
+  }
+
+  removeClientEventListener (eventName) {
+    for (const listener of (this._clientEventListeners[eventName] || [])) {
+      this.rpcClient.off(eventName, listener);
+    }
   }
 
   startConsole (listener) {
     log.debug('Starting to listen for JavaScript console');
-    this.consoleListeners.push(listener);
-    this.rpcClient.on('Console.messageAdded', listener);
-    this.rpcClient.on('Console.messageRepeatCountUpdated', listener);
+    this.addClientEventListener('Console.messageAdded', listener);
+    this.addClientEventListener('Console.messageRepeatCountUpdated', listener);
   }
 
   stopConsole () {
     log.debug('Stopping to listen for JavaScript console');
-    for (const listener of this.consoleListeners) {
-      this.rpcClient.off('Console.messageAdded', listener);
-      this.rpcClient.off('Console.messageRepeatCountUpdated', listener);
-    }
-  }
-
-  /*
-   * Keep track of the network event listeners, so they can be removed
-   */
-  get networkListeners () {
-    this._networkListeners = this._networkListeners || [];
-    return this._networkListeners;
+    this.removeClientEventListener('Console.messageAdded');
+    this.removeClientEventListener('Console.messageRepeatCountUpdated');
   }
 
   startNetwork (listener) {
     log.debug('Starting to listen for network events');
-    this.networkListeners.push(listener);
-    this.rpcClient.on('NetworkEvent', listener);
+    this.addClientEventListener('NetworkEvent', listener);
   }
 
   stopNetwork () {
     log.debug('Stopping to listen for network events');
-    for (const listener of this.networkListeners) {
-      this.rpcClient.off('NetworkEvent', listener);
-    }
+    this.removeClientEventListener('NetworkEvent');
   }
 
   async execute (command, override) {
--- a/lib/remote-debugger.js
+++ b/lib/remote-debugger.js
@@ -116,19 +116,6 @@ class RemoteDebugger extends events.EventEmitter {
     this._navigatingToPage = false;
     this.allowNavigationWithoutReload = false;
 
-    // set up the special callbacks for handling rd events
-    this.specialCbs = {
-      '_rpc_reportIdentifier:': _.noop,
-      '_rpc_forwardGetListing:': this.onPageChange.bind(this),
-      '_rpc_reportConnectedApplicationList:': this.onConnectedApplicationList.bind(this),
-      '_rpc_applicationConnected:': this.onAppConnect.bind(this),
-      '_rpc_applicationDisconnected:': this.onAppDisconnect.bind(this),
-      '_rpc_applicationUpdated:': this.onAppUpdate.bind(this),
-      '_rpc_reportConnectedDriverList:': this.onConnectedDriverList.bind(this),
-      pageLoad: this.pageLoad.bind(this),
-      frameDetached: this.frameDetached.bind(this),
-    };
-
     this.rpcClient = null;
   }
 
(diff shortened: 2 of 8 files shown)


lib/message-handlers.js
```diff
@@ -9,7 +9,7 @@ import _ from 'lodash';
  * These will be added to the prototype.
  */
 
-async function onPageChange (appIdKey, pageDict) {
+async function onPageChange (err, appIdKey, pageDict) {
```
mykola-mokhnach: I thought that the implicit standard is to always put error to the end of the arguments list

lib/remote-debugger.js
```diff
@@ -187,7 +183,7 @@ class RemoteDebugger extends events.EventEmitter {
   }
 
   isConnected () {
```
mykola-mokhnach: does it make sense to also turn it into a getter or this would require to change more client deps?

lib/remote-debugger.js
```diff
@@ -659,34 +656,48 @@ class RemoteDebugger extends events.EventEmitter {
...
+    return this._consoleListeners;
+  }
+
+  startConsole (listener) {
     log.debug('Starting to listen for JavaScript console');
-    if (_.isFunction(fn)) {
-      this.rpcClient.consoleLogEventHandler = fn;
-    }
+    this.consoleListeners.push(listener);
+    this.rpcClient.on('Console.messageAdded', listener);
```
mykola-mokhnach: EventEmitter has an API to retrieve assigned listeners for an event. I would say the `this._consoleListeners` array is redundant

lib/remote-debugger.js
```diff
@@ -659,34 +656,48 @@ class RemoteDebugger extends events.EventEmitter {
...
 
-  async stopConsole () {
+  stopConsole () {
     log.debug('Stopping to listen for JavaScript console');
-    await this.rpcClient.send('Console.disable', {
-      appIdKey: this.appIdKey,
-      pageIdKey: this.pageIdKey,
-    });
+    for (const listener of this.consoleListeners) {
+      this.rpcClient.off('Console.messageAdded', listener);
```
mykola-mokhnach: why not to use https://nodejs.org/api/events.html#events_emitter_removealllisteners_eventname ?

lib/remote-debugger.js
```diff
@@ -659,34 +656,48 @@ class RemoteDebugger extends events.EventEmitter {
...
+    return this._networkListeners;
   }
 
-  async startNetwork (fn) { // eslint-disable-line require-await
+  startNetwork (listener) {
     log.debug('Starting to listen for network events');
-    if (_.isFunction(fn)) {
-      this.rpcClient.networkLogEventHandler = fn;
-    }
+    this.networkListeners.push(listener);
```
mykola-mokhnach: same as above

mykola-mokhnach: also below

lib/rpc-client.js
```diff
@@ -32,7 +32,7 @@ export default class RpcClient {
       bundleId,
       platformVersion = {},
       isSafari = true,
-      specialMessageHandlers = {},
+      // specialMessageHandlers = {},
```
mykola-mokhnach: do we need this comment?

lib/remote-debugger.js
```diff
@@ -159,12 +144,23 @@ class RemoteDebugger extends events.EventEmitter {
...
       socketPath: this.socketPath,
-      specialMessageHandlers: this.specialCbs,
       messageProxy: this.remoteDebugProxy,
       logAllCommunication: this.logAllCommunication,
       logAllCommunicationHexDump: this.logAllCommunicationHexDump,
       fullPageInitialization: this.fullPageInitialization,
     });
+
+    // listen for basic debugger-level events
+    this.rpcClient.on('_rpc_reportSetup:', _.noop);
```
mykola-mokhnach: should we care about listeners cleanup on disconnect to avoid memory leaks?

lib/rpc-client.js
```diff
@@ -231,18 +269,30 @@ export default class RpcClient {
     });
   }
 
+  async connect () { // eslint-disable-line require-await
+    throw new Error(`Sub-classes need to implement a 'connect' function`);
+  }
+
+  async disconnect () { // eslint-disable-line require-await
+    this.messageHandler.removeAllListeners();
```
mykola-mokhnach: 👍

lib/rpc-client.js
```diff
@@ -324,6 +337,7 @@ export default class RpcClient {
       pageIdKey,
     };
 
+    // highlight and then un-highlight the webview
```
mykola-mokhnach: 👍

lib/rpc-client.js
```diff
@@ -231,18 +269,30 @@ export default class RpcClient {
...
   async sendMessage (/* command */) { // eslint-disable-line require-await
     throw new Error(`Sub-classes need to implement a 'sendMessage' function`);
   }
 
-  addTarget (app, targetInfo) {
+  async receive (/* data */) { // eslint-disable-line require-await
+    throw new Error(`Sub-classes need to implement a 'receive' function`);
+  }
+
+  addTarget (err, app, targetInfo) {
```
mykola-mokhnach: should we care about handling `err` arg in these callbacks?

lib/rpc-message-handler.js
```diff
@@ -123,55 +84,65 @@ export default class RpcMessageHandler {
...
-      }
-      this.dataHandlers[msgId](result);
-      this.dataHandlers[msgId] = null;
-    } else if (this.dataHandlers[msgId] === null) {
-      log.error(`Web Inspector returned data for message '${msgId}' ` +
-                `but we already ran that callback! WTF??`);
-    } else {
-      if (msgId || result || error) {
+    if (msgId) {
+      if (this.listeners(msgId).length) {
```
mykola-mokhnach: how about using https://nodejs.org/api/events.html#events_emitter_listenercount_eventname ?

lib/rpc-message-handler.js
```diff
@@ -123,55 +84,65 @@ export default class RpcMessageHandler {
...
+        if (result && result.result && result.result.value) {
```
mykola-mokhnach: I was thinking about enabling [nullish coalescing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) and [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) operators in babel settings. These could allow to simplify such code a lot

lib/rpc-message-handler.js
```diff
@@ -123,55 +84,65 @@ export default class RpcMessageHandler {
...
+        args = [params || params.record];
+        break;
+      case 'Console.messageAdded':
+        args = [params.message];
+        break;
+      case 'Runtime.executionContextCreated':
+        args = [params.context];
+        break;
+      default:
+        // pass
```
mykola-mokhnach: break;

lib/rpc-message-handler.js
```diff
@@ -123,55 +84,65 @@ export default class RpcMessageHandler {
...
+    }
+
+    if (_.isString(method) && method.startsWith('Network.')) {
```
mykola-mokhnach: `_.startsWith` does both checks

lib/rpc-message-handler.js
```diff
@@ -123,55 +84,65 @@ export default class RpcMessageHandler {
...
+      eventNames.push('NetworkEvent');
+      args.push(method);
+    }
+    if (_.isString(method) && method.startsWith('Console.')) {
+      // aggregate Network events, and add original method name to the arguments
+      eventNames.push('ConsoleEvent');
+      args.push(method);
+    }
+
+    if (eventNames.length) {
```
mykola-mokhnach: or (!_.isEmpty(eventNames))

mykola-mokhnach: I think it will never be empty, since we already have at least one element there while initialising the array

lib/rpc-message-handler.js
```diff
@@ -189,12 +160,14 @@ export default class RpcMessageHandler {
       // targetInfo: { targetId: 'page-1', type: 'page' }
       const app = plist.__argument.WIRApplicationIdentifierKey;
       const targetInfo = dataKey.params.targetInfo;
-      await this.specialHandlers.targetCreated(app, targetInfo);
+      // await this.specialHandlers.targetCreated(app, targetInfo);
```
mykola-mokhnach: do we need this comment?

lib/rpc-message-handler.js
```diff
@@ -189,12 +160,14 @@ export default class RpcMessageHandler {
...
+      this.emit('Target.targetCreated', null, app, targetInfo);
       return;
     } else if (method === 'Target.targetDestroyed') {
       const app = plist.__argument.WIRApplicationIdentifierKey;
       const targetInfo = dataKey.params.targetInfo || {targetId: dataKey.params.targetId};
-      await this.specialHandlers.targetDestroyed(app, targetInfo);
+      // await this.specialHandlers.targetDestroyed(app, targetInfo);
```
mykola-mokhnach: do we need this comment?

lib/message-handlers.js
imurchie: No. It's always the first. See https://nodejs.org/en/knowledge/errors/what-are-the-error-conventions/

lib/remote-debugger.js
imurchie: My thinking on this was that the client is exposing event listening, but it's possible that internally the system is also listening. So the only listeners that should be removed are those the client of the system added.

I'll generalize now, and comment.

lib/remote-debugger.js
imurchie: As above. We only want to remove 'external' listeners.

lib/rpc-client.js
imurchie: No.

lib/rpc-message-handler.js
imurchie: 👍

lib/rpc-message-handler.js
imurchie: Also doesn't matter, since for-of will just not have anything to run, and pass through.

lib/remote-debugger.js
```diff
@@ -659,34 +656,42 @@ class RemoteDebugger extends events.EventEmitter {
...
 
-  async startConsole (fn) { // eslint-disable-line require-await
-    log.debug('Starting to listen for JavaScript console');
-    if (_.isFunction(fn)) {
-      this.rpcClient.consoleLogEventHandler = fn;
+  /*
+   * Keep track of the client event listeners so they can be removed
+   */
+  addClientEventListener (eventName, listener) {
+    this._clientEventListeners = this._clientEventListeners || {};
```
mykola-mokhnach: I would rather init it in the constructor

lib/rpc-client.js
```diff
@@ -55,19 +54,55 @@ export default class RpcClient {
...
-    this.specialMessageHandlers.executionContextCreated = this.onExecutionContextAdded.bind(this);
-    this.specialMessageHandlers.garbageCollected = this.onGarbageCollected.bind(this);
-    this.specialMessageHandlers.scriptParsed = this.onScriptParsed.bind(this);
-
     // start with a best guess for the protocol
     this.setCommunicationProtocol(isTargetBased(isSafari, this.platformVersion));
   }
 
+  get contexts () {
+    this._contexts = this._contexts || [];
```
mykola-mokhnach: why not to initialise it in the constructor?

mykola-mokhnach: same below

lib/rpc-message-handler.js
```diff
@@ -233,55 +203,6 @@ export default class RpcMessageHandler {
...
-      },
-      '_rpc_reportConnectedDriverList:': async (plist) => {
-        await this.handleSpecialMessage('_rpc_reportConnectedDriverList:',
-          plist.__argument);
-      },
-      '_rpc_applicationSentData:': async (plist) => {
-        await this.handleDataMessage(plist);
-      },
-    };
+    await this.dispatchDataMessage(msgId, method, params, result, error);
```
mykola-mokhnach: this method seems to have too many arguments

imurchie:
I’ve run this against the driver, with the appropriate modifications. All works.

mykola-mokhnach:
That's cool. Is memory usage also ok? My experience says one has to be really careful with event listener from the perspective of resources management (means all assigned listeners must be released sooner or later) to avoid serious memory leaks

imurchie:
I've tracked all the listeners and all is good, as far as I can tell.

PR #199, opened 2020-02-05
fix: handle async execute on iOS without awaitPromise

imurchie:
On iOS 12.1 MobileSafari within a frame the `window` object gets reset (or we have a different window object?) when going through the atom. But since the code injected is known and stable, and also just retrieving values with no side effects, we can just run it directly.

### DIFF
--- a/lib/remote-debugger.js
+++ b/lib/remote-debugger.js
@@ -422,14 +422,14 @@ class RemoteDebugger extends events.EventEmitter {
     // first create a Promise on the page, saving the resolve/reject functions
     // as properties
     const promiseName = `appiumAsyncExecutePromise${UUID.create().toString().replace(/-/g, '')}`;
-    let script = `var res, rej;` +
-                 `window.${promiseName} = new Promise(function (resolve, reject) {` +
-                 `  res = resolve;` +
-                 `  rej = reject;` +
-                 `});` +
-                 `window.${promiseName}.resolve = res;` +
-                 `window.${promiseName}.reject = rej;` +
-                 `window.${promiseName};`;
+    const script = `var res, rej;` +
+                   `window.${promiseName} = new Promise(function (resolve, reject) {` +
+                   `  res = resolve;` +
+                   `  rej = reject;` +
+                   `});` +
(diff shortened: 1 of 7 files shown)


lib/remote-debugger.js
```diff
@@ -422,14 +422,14 @@ class RemoteDebugger extends events.EventEmitter {
...
     const promiseName = `appiumAsyncExecutePromise${UUID.create().toString().replace(/-/g, '')}`;
-    let script = `var res, rej;` +
-                 `window.${promiseName} = new Promise(function (resolve, reject) {` +
-                 `  res = resolve;` +
-                 `  rej = reject;` +
-                 `});` +
-                 `window.${promiseName}.resolve = res;` +
-                 `window.${promiseName}.reject = rej;` +
-                 `window.${promiseName};`;
+    const script = `var res, rej;` +
```
mykola-mokhnach: I would say quotes also support multiline strings, so pluses are redendant

lib/remote-debugger.js
```diff
@@ -458,20 +458,31 @@ class RemoteDebugger extends events.EventEmitter {
...
+          appIdKey: this.appIdKey,
+          pageIdKey: this.pageIdKey,
+          returnByValue: false,
+        });
+        if (hasValue) {
           // we only put the property on `window` when the callback is called,
           // so if it is there, everything is done
-          return await this.executeAtom('execute_script', [`return window.${promiseName}Value;`, [null, null], subcommandTimeout], frames);
+          return await this.rpcClient.send('Runtime.evaluate', {
+            command: `window.${promiseName}Value;`,
```
mykola-mokhnach: I observe some code duplication here

lib/remote-debugger.js
```diff
@@ -419,41 +419,44 @@ class RemoteDebugger extends events.EventEmitter {
   }
 
   async executeAtomAsync (atom, args, frames) {
+    // helper to send directly to the web inspector
+    const evaluate = async (method, opts) => {
+      return await this.rpcClient.send(method, {
```
mykola-mokhnach: return is not needed

PR #203, opened 2020-02-19
feat: add WIRTypeWebPage as an accepted web page type

imurchie:
iOS 13.4 uses `WIRTypeWebPage` as the web page type, so that must be accepted.

See https://github.com/appium/appium/issues/13948


lib/helpers.js
```diff
@@ -11,6 +11,11 @@ const SAFARI_WEB_VIEW_BUNDLE_ID = 'process-SafariViewService';
 const SAFARI_WEB_VIEW_FULL_BUNDLE_ID = 'com.apple.SafariViewService';
 const WILDCARD_BUNDLE_ID = '*';
 
+const ACCEPTED_PAGE_TYPES = [
+  'WIRTypeWeb',
```
mykola-mokhnach: It would be useful to have a comment there, that this has been added since Xcode 11.4

PR #204, opened 2020-02-19
feat: handle new provisional target creation

imurchie:
On iOS 13.4 there started to be target replacement during the normal flow of events. 

To handle, catch the `Target.didCommitProvisionalTarget` event and keep track, so when the correct target is destroyed, the new target can be used as a replacement.

As with all target handling, there is no page associated with it, so tracking is made more difficult.

### DIFF
--- a/lib/rpc-client.js
+++ b/lib/rpc-client.js
@@ -119,6 +119,7 @@ export default class RpcClient {
 
       // add handlers for internal events
       this.messageHandler.on('Target.targetCreated', this.addTarget.bind(this));
+      this.messageHandler.on('Target.didCommitProvisionalTarget', this.updateTarget.bind(this));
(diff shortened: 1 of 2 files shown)


lib/rpc-client.js
```diff
@@ -312,12 +313,44 @@ export default class RpcClient {
...
+    }
+    // save this, to be used if/when the existing target is destroyed
+    this.targets[app].provisional = {
+      oldTargetId,
+      newTargetId,
+    };
+  }
+
   removeTarget (err, app, targetInfo) {
     if (_.isUndefined(targetInfo) || _.isUndefined(targetInfo.targetId)) {
```
mykola-mokhnach: if (_. isUndefined(targetInfo?.targetId))

PR #208, opened 2020-02-27
fix: remove standard functions from return values

imurchie:
No need to send them back

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -255,7 +255,15 @@ function convertResult (res) {
 
   // with either have an object with a `value` property (even if `null`),
   // or a plain object
-  return _.has(res, 'value') ? res.value : res;
+  let value = _.has(res, 'value') ? res.value : res;
+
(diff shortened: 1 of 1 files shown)


lib/helpers.js
```diff
@@ -255,7 +255,15 @@ function convertResult (res) {
 
   // with either have an object with a `value` property (even if `null`),
   // or a plain object
-  return _.has(res, 'value') ? res.value : res;
+  let value = _.has(res, 'value') ? res.value : res;
+
+  // get rid of noisy functions on objects
+  if (_.isObject(value)) {
+    for (const property of ['ceil', 'clone', 'floor', 'round', 'scale', 'toString']) {
```
mykola-mokhnach: are you sure `toString` is not needed?

imurchie: Yes. They are functions from the context of the JS running in Safari. At this point they are strings, and never get de-serialized.

PR #210, opened 2020-03-03
chore: retry for client connection in case of failure

yashLadha:


yashLadha:
@mykola-mokhnach  The purpose of this PR is to retry on command in the case at startup, RPC client is not connected. Juts a safety check to ensure as we are doing retries at various parts in appium and this practice can be seen in other frameworks as well.

lib/remote-debugger.js
```diff
@@ -28,6 +28,7 @@ const SELECT_APP_RETRY_SLEEP_MS = 500;
 const REMOTE_DEBUGGER_PORT = 27753;
 const SAFARI_BUNDLE_ID = 'com.apple.mobilesafari';
 const BLANK_PAGE_URL = 'about:blank';
+const WDA_CONNECT_RETRIES = 1;
```
imurchie: What is `WDA` here?

lib/remote-debugger.js
```diff
@@ -406,9 +407,22 @@ class RemoteDebugger extends events.EventEmitter {
     log.debug(`Selected page after ${timer.getDuration().asMilliSeconds.toFixed(0)}ms`);
   }
 
+  async retryConnector () {
+    for (let retryCount = 0; retryCount < WDA_CONNECT_RETRIES; ++retryCount) {
+      await this.connect(1);
```
imurchie: Why a timeout of 1ms?

yashLadha: To basically halt for a minimal time, as `0` is also not executed immediately in some cases.

mykola-mokhnach:
> The purpose of this PR is to retry on command in the case at startup, RPC client is not connected. Juts a safety check to ensure as we are doing retries at various parts in appium and this practice can be seen in other frameworks as well.

Does it really help? Could you please provide some log where the retry really helps to recover from an error? In my experience retries could make the code really slow, especially if they are used in frequently called areas. Perhaps, the original issue could (should) be fixed in some other way?

yashLadha:
Closing for now.

PR #211, opened 2020-03-09
refactor: use gulp tasks instead of install scripts for atoms

imurchie:
No need to have two different build systems, one of which is used every once in a while. Move to `gulp` to replace the current `make`/`bash` system.


gulpfile.js
```diff
@@ -2,6 +2,26 @@
...
+const del = require('del');
+const path = require('path');
+const log = require('fancy-log');
+const B = require('bluebird');
+
+
+const SELENIUM_BRANCH = 'selenium-3.141.59';
+const SELENIUM_GITHUB = 'https://github.com/SeleniumHQ/selenium.git';
+
+const TMP_DIRECTORY = path.resolve(__dirname, 'tmp');
```
mykola-mokhnach: shall the system temp dir be used instead?

gulpfile.js
```diff
@@ -16,3 +36,120 @@ boilerplate({
     ],
   }
 });
+
+gulp.task('selenium:mkdir', function seleniumMkdir () {
+  log(`Creating '${TMP_DIRECTORY}'`);
+  return mkdirp(TMP_DIRECTORY);
```
mykola-mokhnach: is this synchronous call?

imurchie: No, `gulp` will wait for any promise that is returned to it.

gulpfile.js
imurchie: We often need to do things to the bare Selenium atoms that are downloaded, so I'd rather it exist locally.

PR #213, opened 2020-03-13
chore: use uuid from appium-support

imurchie:


mykola-mokhnach:
https://travis-ci.org/github/appium/appium-xcuitest-driver/jobs/662092796#L7869 - it looks like something got broken after the recent wd update

PR #215, opened 2020-03-16
refactor: easier isActive logic

imurchie:


lib/remote-debugger.js
```diff
@@ -370,7 +368,7 @@ class RemoteDebugger extends events.EventEmitter {
 
   searchForPage (appsDict, currentUrl = null, ignoreAboutBlankUrl = false) {
     for (const appDict of _.values(appsDict)) {
-      if (!appDict || appDict.isActive === ACTIVE_APP_CODE || !appDict.pageArray || appDict.pageArray.promise) {
+      if (!appDict || !appDict.isActive || !appDict.pageArray || appDict.pageArray.promise) {
```
mykola-mokhnach: we could use conditional chaining here

PR #216, opened 2020-03-16
refactor: clean up parameter error handling

imurchie:
Easier to just throw.

Plus, a simple benchmark shows that `try`/`catch` for control flow is very inefficient
```
[12:51:44] assert/catch x 36,599 ops/sec ±8.25% (60 runs sampled)
[12:51:51] if/then x 164,072,843 ops/sec ±1.51% (82 runs sampled)
[12:51:51] Fastest is if/then
```

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -161,12 +161,10 @@ function getPossibleDebuggerAppKeys (bundleIds, appDict) {
 }
 
 function checkParams (params) {
-  let errors = [];
-  for (const [param, value] of _.toPairs(params)) {
-    if (_.isNil(value)) {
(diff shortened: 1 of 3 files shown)


lib/helpers.js
```diff
@@ -163,14 +163,12 @@ function getPossibleDebuggerAppKeys (bundleIds, appDict) {
 function checkParams (params) {
   let errors = [];
```
mykola-mokhnach:
```
const errors = _.toPairs(params).filter(([param, value]) => _.isNil(value)).map(([param, value]) => param);
```

PR #225, opened 2020-04-13
refactor: move protocol out of rpc generation

imurchie:
Separate the protocol messages from the RPC generation, which is iOS specific. I've been working on integrating Chrome and this is a first step. Also makes the protocol easier to handle.

### DIFF
--- a/lib/protocol/index.js
+++ b/lib/protocol/index.js
@@ -123,8 +123,8 @@ function getProtocolCommand (id, method, opts = {}) {
     throw new Error(`Unknown command: '${method}'`);
   }
 
-  const params = paramNames.reduce(function (params, param) {
-    params[param] = opts[param];
+  const params = paramNames.reduce(function (params, name) {
+    params[name] = opts[name];
     return params;
   }, {});
   return getCommand(id, method, params);
--- a/lib/protocol/index.js
+++ b/lib/protocol/index.js
@@ -1,5 +1,6 @@
 const OBJECT_GROUP = 'console';
 
+// See https://github.com/WebKit/webkit/tree/master/Source/JavaScriptCore/inspector/protocol
 const COMMANDS = {
   //#region APPLICATIONCACHE DOMAIN
   'ApplicationCache.enable': [],
--- a/lib/protocol/index.js
+++ b/lib/protocol/index.js
@@ -1,0 +1,114 @@
+const OBJECT_GROUP = 'console';
+
+const COMMANDS = {
+  /* APPLICATIONCACHE DOMAIN */
+  'ApplicationCache.enable': [],
+  'ApplicationCache.getFramesWithManifests': [],
+
+  /* CANVAS DOMAIN */
+  'Canvas.enable': [],
+
+  /* CONSOLE DOMAIN */
+  'Console.disable': [],
+  'Console.enable': [],
+  'Console.getLoggingChannels': [],
+  'Console.setLoggingChannelLevel': ['source', 'level'],
+
+  /* CSS DOMAIN */
+  'CSS.enable': [],
+
+  /* DATABASE DOMAIN */
+  'Database.enable': [],
+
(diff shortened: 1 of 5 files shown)


lib/protocol/index.js
```diff
@@ -0,0 +1,114 @@
+const OBJECT_GROUP = 'console';
+
+const COMMANDS = {
```
mykola-mokhnach, line 3: I usually also add a link to the official doc for such declarations

lib/protocol/index.js
```diff
@@ -0,0 +1,114 @@
...
+  /* APPLICATIONCACHE DOMAIN */
```
mykola-mokhnach, line 4: we could use the `region` notation to make sections easily collapsable in IDEs: https://code.visualstudio.com/updates/v1_17#_folding-regions

lib/protocol/index.js
```diff
@@ -0,0 +1,114 @@
...
+
+  /* WORKER DOMAIN */
+  'Worker.enable': [],
+};
+
+function getCommand (id, method, params = {}) {
+  return {
+    id,
+    method,
+    params: Object.assign({
```
mykola-mokhnach, line 89: using `Object.assign` will override default properties even if these contain null/undefined in `params`. Should we use lodash's `_.defaults` instead?

lib/protocol/index.js
```diff
@@ -0,0 +1,114 @@
...
+}
+
+function getProtocolCommand (id, method, opts = {}) {
+  const paramNames = COMMANDS[method];
+  if (!paramNames) {
+    throw new Error(`Unknown command: '${method}'`);
+  }
+
+  const params = paramNames.reduce(function (params, param) {
+    params[param] = opts[param];
```
mykola-mokhnach, line 107: nit: it's a bit confusing to read if the result name is the same as the accumulator name. I'd rather rename the first arg of the reducer function to `acc` or `result`

lib/rpc/remote-messages.js
```diff
@@ -1,8 +1,29 @@
 import _ from 'lodash';
+import getProtocolCommand from '../protocol';
 
 
 const OBJECT_GROUP = 'console';
 
+const MINIMAL_COMMAND = 'getMinimalCommand';
+const FUll_COMMAND = 'getFullCommand';
```
mykola-mokhnach, line 8: perhaps `FULL_COMMAND`?

lib/rpc/remote-messages.js
```diff
@@ -41,8 +60,7 @@ class RemoteMessages {
     };
   }
 
-  setSenderKey (opts = {}) {
-    const {connId, senderId, appIdKey, pageIdKey} = opts;
+  setSenderKey (connId, senderId, appIdKey, pageIdKey) {
```
mykola-mokhnach, line 63: I liked the previous implementation more. For a client it's quite easier to mix arguments between each other if the count of these is greater than two

lib/rpc/rpc-client.js
```diff
@@ -460,7 +460,9 @@ export default class RpcClient {
     await this.send('Memory.enable', sendOpts, false);
     await this.send('ApplicationCache.enable', sendOpts, false);
     await this.send('ApplicationCache.getFramesWithManifests', sendOpts, false);
-    await this.send('Timeline.setInstruments', sendOpts, false);
+    await this.send('Timeline.setInstruments', Object.assign({
```
mykola-mokhnach, line 463: same comment as above

lib/rpc/remote-messages.js
imurchie: This is only called in one place, below. The overhead of creating a new object, etc, is not needed.

lib/rpc/rpc-client.js
imurchie: This involves nothing that is inputed from anywhere, so the caution is unnecessary.

PR #234, opened 2020-06-22
feat: add webInspectorMaxFrameLength option

KazuCocoa:
https://github.com/appium/appium-ios-device/pull/73 and https://github.com/appium/appium/issues/14700

### DIFF
--- a/lib/remote-debugger.js
+++ b/lib/remote-debugger.js
@@ -92,8 +92,9 @@ class RemoteDebugger extends EventEmitter {
     this.logAllCommunicationHexDump = logAllCommunicationHexDump;
     this.socketChunkSize = socketChunkSize;
 
-    // When the value is 'undefined', the web inspector max frame length will be the default value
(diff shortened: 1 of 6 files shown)


mykola-mokhnach:
I would say this would be just a hacky workaround, that does not solve the original problem. And the original problem is that the internal data buffer created by Node in appium-ios-device gets overflown, while the reason for it stays unknown.

lib/remote-debugger.js
```diff
@@ -89,6 +92,9 @@ class RemoteDebugger extends EventEmitter {
     this.logAllCommunicationHexDump = logAllCommunicationHexDump;
     this.socketChunkSize = socketChunkSize;
 
+    // When the value is 'undefined', the web inspector max frame length will be the default value
```
mykola-mokhnach, line 95:
```
if (_.isInteger(webInspectorMaxFrameLength)) {
  this.webInspectorMaxFrameLength = webInspectorMaxFrameLength
}
```

mykola-mokhnach (approved):

PR #245, opened 2020-11-21
chore: use Azure pipelines and bump iOS versions

dpgraham:


ci-jobs/pr-validation.yml
```diff
@@ -0,0 +1,25 @@
...
+- name: SHOW_XCODE_LOG
+  value: true
+
+stages:
+- stage: Unit_Tests
+  jobs:
+  - template: ./templates/node-e2e-template.yml
+- stage: Integration_Tests
+  jobs:
+  - template: ./templates/remote-debugger-e2e-template.yml
```
mykola-mokhnach, line 12: This could be implemented in a loop to avoid unnecessary duplication. See https://github.com/appium/WebDriverAgent/blob/c7804d9cec911ceefce555d31d2a3b6d3d733e94/.azure-pipelines.yml#L333 for example

ci-jobs/templates/ios-e2e-template.yml
```diff
@@ -0,0 +1,47 @@
...
+      displayName: Xcode Select ${{ parameters.xcodeVersion }}
+    - script: xcodebuild -version
+      displayName: Log Xcode Version
+    - script: xcrun simctl list
+      displayName: List Installed Simulators
+    - script: xcrun simctl list runtimes
+      displayName: List Installed Runtimes
+    - task: NodeTool@0
+      inputs:
+        versionSpec: 12.x
```
mykola-mokhnach, line 39: It makes sense to move node version to global variables, so it would be easy to bump it when necessary

mykola-mokhnach (approved):

PR #267, opened 2022-07-02
fix: build atoms fail when inject

li-zhixin:


gulpfile.js
```diff
@@ -81,8 +81,8 @@ gulp.task('atoms:mkdir', function atomsMkdir () {
 
 gulp.task('atoms:inject', function atomsInject () {
   log('Injecting build file into Selenium build');
-  return fs.promises.copyFile(
```
mykola-mokhnach, line 84: what is the actual problem with this call?

package.json
```diff
@@ -64,23 +64,24 @@
...
     "@appium/eslint-config-appium": "^6.0.0",
+    "@appium/gulp-plugins": "^7.0.0",
     "@semantic-release/git": "^10.0.1",
     "appium-ios-simulator": "^4.0.0",
     "chai": "^4.1.2",
     "chai-as-promised": "^7.1.1",
     "eslint-config-prettier": "^8.5.0",
     "fancy-log": "^2.0.0",
     "finalhandler": "^1.1.2",
+    "fs-extra": "^10.1.0",
```
mykola-mokhnach, line 76: do we really need to import an additional module to just copy a single file?

gulpfile.js
li-zhixin: Here you actually expect the folder `atoms_build_dir` to be copied to `. /tmp/selenium/javascript/appium-atoms-driver, but instead the method to copy the file is being called.

mykola-mokhnach: thanks, I get it.

Usually I'm not very into adding additional imports when it is not really necessary. Would it be possible to add a custom implementation of async recursive folder copy?

li-zhixin: I see, how about now?

gulpfile.js
```diff
@@ -23,6 +23,26 @@ const TEMP_BUILD_DIRECTORY_NAME = 'appium-atoms-driver';
 
 const ATOMS_BUILD_TARGET = 'build_atoms';
 
+const copyFolderRecursive = async (src, dest) => {
```
mykola-mokhnach, line 26: this could be a vanilla function. There is no point in making it anonymous

gulpfile.js
```diff
@@ -23,6 +23,26 @@ const TEMP_BUILD_DIRECTORY_NAME = 'appium-atoms-driver';
...
+  const entries = await fs.promises.readdir(src, { withFileTypes: true });
+  try {
+    await fs.promises.access(dest, fs.constants.R_OK);
+  } catch (err) {
+    await fs.promises.mkdir(dest);
```
mykola-mokhnach, line 31: please add an option to make the dir creation recursive

gulpfile.js
```diff
@@ -23,6 +23,26 @@ const TEMP_BUILD_DIRECTORY_NAME = 'appium-atoms-driver';
...
+  }
+  for (let entry of entries) {
```
mykola-mokhnach, line 33: let -> const

gulpfile.js
```diff
@@ -23,6 +23,26 @@ const TEMP_BUILD_DIRECTORY_NAME = 'appium-atoms-driver';
...
+    const srcPath = path.join(src, entry.name);
+    const destPath = path.join(dest, entry.name);
+    if (entry.isDirectory()) {
+      await copyFolderRecursive(srcPath, destPath);
+    } else if (entry.isFile()) {
+      await fs.promises.copyFile(srcPath, destPath);
+    } else {
+      throw new Error(`Cannot copy: ${srcPath}`);
```
mykola-mokhnach, line 41: I'd say we simply skip an entry if it is not a file or directory

gulpfile.js
```diff
@@ -81,8 +101,8 @@ gulp.task('atoms:mkdir', function atomsMkdir () {
 
 gulp.task('atoms:inject', function atomsInject () {
   log('Injecting build file into Selenium build');
-  return fs.promises.copyFile(
-    ATOMS_BUILD_DIRECTORY, `${SELENIUM_DIRECTORY}/javascript/${TEMP_BUILD_DIRECTORY_NAME}`
+  return copyFolderRecursive(
+      ATOMS_BUILD_DIRECTORY, `${SELENIUM_DIRECTORY}/javascript/${TEMP_BUILD_DIRECTORY_NAME}`
```
mykola-mokhnach, line 105: we could use `path.join` for the merge

mykola-mokhnach (approved):

mykola-mokhnach:
Thanks @li-zhixin 
could you please sign CLA, so we can merge the PR?

li-zhixin:
> Thanks @li-zhixin could you please sign CLA, so we can merge the PR?

done

gulpfile.js
li-zhixin: Do you mean like this?
``` javascript
  return copyFolderRecursive(
    ATOMS_BUILD_DIRECTORY, path.join(SELENIUM_DIRECTORY, 'javascript', TEMP_BUILD_DIRECTORY_NAME)
  );
```

mykola-mokhnach: yes

jlipps:
I am pretty sure our whole atoms build system is broken with the move in Selenium to Bazel?

PR #268, opened 2022-07-13
feat: update selenium atoms based on selenium 4.11.0

jlipps:
WIP of bringing the newest selenium atoms over. things don't work and it's not exactly clear why. debugging is required.


KazuCocoa:
@jlipps Do you remember where this atoms change come from?

This repo's `build:atoms` command generates atoms based on selenium `3.141.59`, which was in 2018. Then, this might not help to update atoms...

`selenium-webdriver-4.6.1` Ruby client only had below files in `lib/selenium/webdriver/atoms`.

```
findElements.js     getAttribute.js     isDisplayed.js      mutationListener.js
```
https://github.com/SeleniumHQ/selenium/blob/selenium-4.6.0/rb/BUILD.bazel

Python has only find_elements, get_attribute, isdisplayed.
https://github.com/SeleniumHQ/selenium/blob/selenium-4.6.0/py/BUILD.bazel

so... potentially we only need these atoms...?
(My understanding might be wrong though)

jlipps:
@KazuCocoa yeah maybe it was 3.141.59, since i think maybe that was what was in master when i did this? i don't remember.

are you saying that many atoms have been removed in favour of direct remote debugger calls?

KazuCocoa:
> are you saying that many atoms have been removed in favour of direct remote debugger calls?

I may not properly understand this usage well yet. I initially thought old(?) Selenium clients also used the same atoms, so this remote-debugger referred to them from the selenium project, although the latest selenium client (I only checked Ruby and python tho), had a few atoms only. Others were just endpoint calls defined in W3C.

So... I just wondered if we only needed the same a few atoms only, not others nowadays (we may also need to update some this repo's implementation though, I guess). Sorry, this is not so concrete question now.

jlipps:
Hmm no worries. Yeah maybe we don't need all the same atoms as before. But we should probably support the same set of behaviours. Maybe more of the clients just send their own JS snippets for many of these commands now, so we can get away with fewer atoms and just Execute Script?

jlipps:
@mykola-mokhnach @KazuCocoa this PR has new atoms from selenium 4.11. it also has one that i built 'manually' but hope to add as a PR to selenium. all the remote debugger e2e tests pass, and the same tests pass in the xcuitest driver web set that passed before (not all of them passed initially for me on my machine, at least on ios 16.2, idk why). all that to say, this is now ready for review!

hopefully an atoms update will fix some of the issues that our users have been reporting.

KazuCocoa:
Probably can we remove https://github.com/appium/appium-remote-debugger/blob/master/atoms-notes.md for now..?

KazuCocoa (approved):
I can't check everything, but your comment might indicate this will work (at least some points)

scripts/common.js
```diff
@@ -92,104 +74,79 @@ async function atomsCleanDir () {
...
-        return resolve(code);
-      } else {
-        return reject(code);
-      }
-    });
-    proc.start();
-  });
+async function getBazelOutDir () {
+  log(`Finding bazel output dir`);
+  const relativeDir = (await B.promisify(glob)('*-fastbuild', {cwd: BAZEL_OUT_BASEDIR}))[0];
```
mykola-mokhnach, line 87: what glob version do we import? I believe the last version of it is async

scripts/common.js
```diff
@@ -92,104 +74,79 @@ async function atomsCleanDir () {
...
-    }
-    const filename = path.basename(p);
-    if (['_exports.js', '_ie.js', '_build_atoms.js', 'deps.js'].some((x) => filename.endsWith(x))) {
-      return false;
-    }
-    return true;
-  };
+async function atomsBuild () {
+  log(`Running bazel build for ${BAZEL_ATOMS_TARGET}`);
+  await exec('bazel', ['build', BAZEL_ATOMS_TARGET], {cwd: SELENIUM_DIRECTORY});
```
mykola-mokhnach, line 93: can these operations be executed in parallel or they depend on each other?

scripts/common.js
```diff
@@ -92,104 +74,79 @@ async function atomsCleanDir () {
...
+  log(`Running bazel build for ${BAZEL_SE_ATOMS_TARGET}`);
+  await exec('bazel', ['build', BAZEL_SE_ATOMS_TARGET], {cwd: SELENIUM_DIRECTORY});
+  log(`Bazel builds complete`);
+}
 
-  const filesToCopy = (await (B.promisify(glob)('**/*.js', {
+async function atomsCopyAtoms (atomsDir, fileFilter) {
+  log(`Copying any atoms found in ${atomsDir} to atoms dir`);
+  if (!fileFilter) {
+    fileFilter = () => true;
```
mykola-mokhnach, line 104: wouldn't it be more readable to assign the default value above in the args definition?

scripts/common.js
```diff
@@ -92,104 +74,79 @@ async function atomsCleanDir () {
...
 }
 
 async function atomsTimestamp () {
+  log(`Adding timestamp to atoms build dir`);
   const {stdout} = await exec('git', ['log', '-n', '1', '--decorate=full'], {cwd: SELENIUM_DIRECTORY});
   await fs.promises.writeFile(LAST_UPDATE_FILE, Buffer.from(`${new Date()}\n\n${stdout}`));
 }
 
-module.exports.importAtoms = async function importAtoms() {
+module.exports.importAtoms = async function importAtoms(runBazelClean) {
```
mykola-mokhnach, line 135: not sure if we need to explicitly mention bazel. `shouldClean` would be more universal

scripts/common.js
jlipps: i'll try to update all dependencies as part of this PR

lib/rpc/rpc-message-handler.js
```diff
@@ -110,7 +110,7 @@ export default class RpcMessageHandler extends EventEmitters {
     switch (method) {
       case 'Page.frameStoppedLoading':
         eventNames.push('Page.frameNavigated');
-        // eslint-disable-line no-fallthrough
```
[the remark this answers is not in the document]
mykola-mokhnach: I don't know. I assume it is expected if this comment is present there

scripts/common.js
```diff
@@ -1,45 +1,26 @@
...
 
+const BAZEL_WD_ATOMS_TARGET = '//javascript/webdriver/atoms/...';
+const BAZEL_ATOMS_TARGET = '//javascript/atoms/...';
+const BAZEL_SE_ATOMS_TARGET = '//javascript/selenium-atoms/...';
+
 const WORKING_ROOT_DIR = path.resolve(__dirname, '..');
 const TMP_DIRECTORY = path.resolve(WORKING_ROOT_DIR, 'tmp');
 const SELENIUM_DIRECTORY = path.resolve(TMP_DIRECTORY, 'selenium');
+const BAZEL_OUT_BASEDIR = path.resolve(SELENIUM_DIRECTORY, 'bazel-out');
+const BAZEL_FRAGMENTS_DIR = path.join('bin', 'javascript', 'atoms', 'fragments');
```
mykola-mokhnach, line 18: we could extract the path to 'bin', 'javascript' to a separate variable

scripts/common.js
```diff
@@ -92,104 +73,76 @@ async function atomsCleanDir () {
 
 async function atomsClean () {
   log('Building atoms');
-  await exec('./go', ['clean'], {cwd: SELENIUM_DIRECTORY});
+  await exec('bazel', ['clean'], {cwd: SELENIUM_DIRECTORY});
```
mykola-mokhnach, line 76: should there be some validation about bazel is installed on the machine?

scripts/common.js
```diff
@@ -92,104 +73,76 @@ async function atomsCleanDir () {
...
-        return resolve(code);
-      } else {
-        return reject(code);
-      }
-    });
-    proc.start();
-  });
+async function getBazelOutDir () {
+  log(`Finding bazel output dir`);
+  const relativeDir = (await glob('*-fastbuild', {cwd: BAZEL_OUT_BASEDIR}))[0];
```
mykola-mokhnach, line 86: I assume there should be some validation in case no matches are found

scripts/common.js
```diff
@@ -92,104 +73,76 @@ async function atomsCleanDir () {
...
-      return false;
-    }
-    const filename = path.basename(p);
-    if (['_exports.js', '_ie.js', '_build_atoms.js', 'deps.js'].some((x) => filename.endsWith(x))) {
-      return false;
-    }
-    return true;
-  };
+async function atomsBuild () {
+  log(`Running bazel build for ${BAZEL_ATOMS_TARGET}`);
```
mykola-mokhnach, line 91: I assume it still makes sense to put these calls into a loop because multiple args are the same

lib/rpc/rpc-message-handler.js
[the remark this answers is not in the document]
jlipps: logically the fallthrough can only go to `default` which is just `break` itself. i bet it is left over from a time when the fallthrough mattered.

scripts/common.js
jlipps: if it's not, we will get a ENOENT error, which is easily understandable by the people running this script (not user facing, this is only us that will run this script). i would rather just note the requirements in the README than have custom errors only several people will ever potentially see.

lib/rpc/rpc-message-handler.js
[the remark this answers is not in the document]
mykola-mokhnach: in this particular case the fallthrough goes to `args = [`'${method}' event`];`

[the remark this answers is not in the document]
jlipps: ah yes you're right. will fix.

mykola-mokhnach (approved):

scripts/common.js
```diff
@@ -1,45 +1,27 @@
...
 
+const BAZEL_WD_ATOMS_TARGET = '//javascript/webdriver/atoms/...';
+const BAZEL_ATOMS_TARGET = '//javascript/atoms/...';
+const BAZEL_SE_ATOMS_TARGET = '//javascript/selenium-atoms/...';
+
 const WORKING_ROOT_DIR = path.resolve(__dirname, '..');
 const TMP_DIRECTORY = path.resolve(WORKING_ROOT_DIR, 'tmp');
 const SELENIUM_DIRECTORY = path.resolve(TMP_DIRECTORY, 'selenium');
+const BAZEL_OUT_BASEDIR = path.resolve(SELENIUM_DIRECTORY, 'bazel-out');
+const JS_RELATIVE_DIR = path.join('bin', 'javascript');
```
mykola-mokhnach, line 18: I prefer to use absolute paths everywhere. You could use `WORKING_ROOT_DIR` there

test/functional/safari-e2e-specs.js
```diff
@@ -13,8 +13,8 @@ chai.should();
 chai.use(chaiAsPromised);
 
 const SIM_NAME = process.env.SIM_DEVICE_NAME || `appium-test-${util.uuidV4()}`;
-const DEVICE_NAME = process.env.DEVICE_NAME || 'iPhone 6';
-const PLATFORM_VERSION = process.env.PLATFORM_VERSION || '12.1';
+const DEVICE_NAME = process.env.DEVICE_NAME || 'iPhone 14';
```
mykola-mokhnach, line 16: should we also eventually bump versions in the CI scripts?

scripts/common.js
jlipps: this is not possible to know in advance, since this path will be relative to an architecture-dependent dir we need to find as a result of glob.

test/functional/safari-e2e-specs.js
jlipps: yes, we can take a look at that

jlipps:
OK I have published 10.0.0-beta.0. I did not merge this PR yet since autorelease is turned on.

PR #274, opened 2022-11-26
chore: update protocol a few and add links to each protocol

KazuCocoa:
Update protocol from https://github.com/WebKit/webkit/tree/main/Source/JavaScriptCore/inspector/protocol

This does not change anything for existing commands since no new command is called. I have just added links to each protocol and add `disable` command and a few which might be interested in.

### DIFF
--- a/lib/protocol/index.js
+++ b/lib/protocol/index.js
@@ -2,32 +2,66 @@ const OBJECT_GROUP = 'console';
 
 // See https://github.com/WebKit/webkit/tree/master/Source/JavaScriptCore/inspector/protocol
 const COMMANDS = {
+  // https://github.com/WebKit/WebKit/blob/main/Source/JavaScriptCore/inspector/protocol/Animation.json
+  //#region WORKER DOMAIN
+  'Animation.enable': [], // Enables Canvas domain events
+  'Animation.disable': [], // Disables Canvas domain events
+  //#endregion
+
+  // https://github.com/WebKit/WebKit/blob/main/Source/JavaScriptCore/inspector/protocol/ApplicationCache.json
   //#region APPLICATIONCACHE DOMAIN
   'ApplicationCache.enable': [],
+  'ApplicationCache.disable': [],
   'ApplicationCache.getFramesWithManifests': [],
   //#endregion
 
-  //#region CANVAS DOMAIN
-  'Canvas.enable': [],
+  // https://github.com/WebKit/WebKit/blob/main/Source/JavaScriptCore/inspector/protocol/Audit.json
+  //#region APPLICATIONCACHE DOMAIN
   //#endregion
 
+  // https://github.com/WebKit/WebKit/blob/main/Source/JavaScriptCore/inspector/protocol/Browser.json
+  //#region APPLICATIONCACHE DOMAIN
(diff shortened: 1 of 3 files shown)


mykola-mokhnach:
Thanks @KazuCocoa for looking into it. I hope you'd be finally able to document this stuff better.
I would be happy to understand:
- how to inspect/sniff actual protocol commands and add/edit them in this module
- how the protocol works in general (data models, set of supported commands, etc)

lib/protocol/index.js
```diff
@@ -2,103 +2,183 @@ const OBJECT_GROUP = 'console';
...
   'Debugger.setAsyncStackTraceDepth': ['depth'],
   'Debugger.setBreakpointsActive': ['active'],
   'Debugger.setPauseForInternalScripts': ['shouldPause'],
   'Debugger.setPauseOnAssertions': ['enabled'],
   'Debugger.setPauseOnExceptions': ['state'],
   //#endregion
 
+  // https://github.com/WebKit/WebKit/blob/main/Source/JavaScriptCore/inspector/protocol/DOM.json
   //#region DOM DOMAIN
-  'DOM.getDocument': [],
```
mykola-mokhnach, line 39: was this command removed?

KazuCocoa: no, that was my mistake for domdebugger..

KazuCocoa:
I just read https://github.com/WebKit/WebKit/blob/main/Source/JavaScriptCore/inspector/protocol and pasted some of them (and tweaked the order following the WebKit's definitions. Haven't checked each behavior well yet, but basically i modified or send a new command in https://github.com/appium/appium-remote-debugger/blob/05b686e8be80659d4abbd6c499287b41699d629a/lib/rpc/rpc-client.js#L455-L500 (mostly a real device in my case) with enabling https://github.com/appium/appium-remote-debugger/blob/05b686e8be80659d4abbd6c499287b41699d629a/lib/rpc/rpc-client.js#L39 .

(Original motivation was to find clearing the local storage stuff, but no good one though)

Each behavior is not the same, but similar to https://chromedevtools.github.io/devtools-protocol/. 


https://github.com/appium/appium-remote-debugger/blob/05b686e8be80659d4abbd6c499287b41699d629a/development-notes.md also helps for this. As the document, I also try to estimate each behavior based on https://chromedevtools.github.io/devtools-protocol/. Basically Devtools protocol has much features, so WebKit one is kind of subset.

package.json
```diff
@@ -47,7 +47,7 @@
     "teen_process": "^2.0.0"
   },
   "scripts": {
-    "build": "rm -rf build && babel --out-dir=build/lib lib && babel --out-dir=build index.js",
+    "build": "rm -rf build && babel --out-dir=build/lib lib && babel --out-dir=build index.js && babel --out-dir=build/bin bin",
```
KazuCocoa, line 50: then, `inspect-safari` for a simulator may work. Potentially we need to tweak the bin additionally as another PR.

KazuCocoa:
https://github.com/appium/appium-xcuitest-driver/pull/1455 is one way to inspect communications, and try some commands via appium

mykola-mokhnach (approved):

PR #276, opened 2022-11-28
fix: fix scripts for atoms

KazuCocoa:
Fixed scripts associated with atoms as minimal way so that `npm run build:atoms` can generate atoms as same as previously.

This won't include more changes in `scripts` for now to keep the change minimal.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,9 +57,9 @@
     "test": "mocha --exit --timeout 1m \"./test/unit/**/*-specs.js\"",
     "e2e-test": "mocha --exit --timeout 1m \"./test/functional/**/*-specs.js\"",
     "inspect-safari": "node build/bin/web_inspector_proxy.js",
-    "build:atoms": "babel --out-dir=build/scripts scripts && node build/scripts/build-atoms.js",
-    "build:atoms:import": "babel --out-dir=build/scripts scripts && node build/scripts/import-atoms.js",
-    "build:selenium": "babel --out-dir=build/scripts scripts && node build/scripts/build-selenium.js"
+    "build:atoms": "node scripts/build-atoms.js",
+    "build:atoms:import": "node scripts/import-atoms.js",
+    "build:selenium": "node scripts/build-selenium.js"
   },
   "pre-commit": [
     "precommit-msg",
--- a/package.json
+++ b/package.json
@@ -57,9 +57,9 @@
     "test": "mocha --exit --timeout 1m \"./test/unit/**/*-specs.js\"",
     "e2e-test": "mocha --exit --timeout 1m \"./test/functional/**/*-specs.js\"",
     "inspect-safari": "node build/bin/web_inspector_proxy.js",
-    "build:atoms": "node scripts/build-atoms.js",
-    "build:atoms:import": "node scripts/import-atoms.js",
-    "build:selenium": "node scripts/build-selenium.js"
+    "build:atoms": "babel --out-dir=build/scripts scripts && node build/scripts/build-atoms.js",
+    "build:atoms:import": "babel --out-dir=build/scripts scripts && node build/scripts/import-atoms.js",
+    "build:selenium": "babel --out-dir=build/scripts scripts && node build/scripts/build-selenium.js"
   },
   "pre-commit": [
     "precommit-msg",
--- a/scripts/common.js
+++ b/scripts/common.js
@@ -8,7 +8,7 @@ const B = require('bluebird');
 const SELENIUM_BRANCH = 'selenium-3.141.59';
 const SELENIUM_GITHUB = 'https://github.com/SeleniumHQ/selenium.git';
 
-const WORKING_ROOT_DIR = path.resolve(__dirname, '..', '..');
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -57,9 +57,9 @@
     "test": "mocha --exit --timeout 1m \"./test/unit/**/*-specs.js\"",
     "e2e-test": "mocha --exit --timeout 1m \"./test/functional/**/*-specs.js\"",
     "inspect-safari": "node build/bin/web_inspector_proxy.js",
-    "build:atoms": "node scripts/build-atoms.js",
-    "build:atoms:import": "node scripts/import-atoms.js",
-    "build:selenium": "node scripts/build-selenium.js"
+    "build:atoms": "babel --out-dir=build/scripts scripts && node build/scripts/build-atoms.js",
```
mykola-mokhnach, line 60: scripts should be buildable without babel, shouldn't they?

scripts/common.js
```diff
@@ -166,11 +167,12 @@ async function atomsCopy () {
     return true;
   };
 
-  const filesToCopy = (await glob('**/*.js', {
+  const copyTarget = glob.sync('**/*.js', {
```
mykola-mokhnach, line 170: why the sync method should be used?

scripts/common.js
```diff
@@ -8,10 +8,11 @@ const B = require('bluebird');
 const SELENIUM_BRANCH = 'selenium-3.141.59';
```
mykola-mokhnach, line 8: shall we always use the same branch or change it to master/main?

KazuCocoa: This is maybe tag https://github.com/SeleniumHQ/selenium/releases/tag/selenium-3.141.59
Selenium 4.x uses bazel so this build steps won't work over the 3.141.59. We can update naming as well when we update the building step for selenium 4's atoms

package.json
KazuCocoa: i could be wrong.

Current script got the below syntax error. When I have added `type`: `module` like https://blog.logrocket.com/you-dont-need-babel-with-node/ .. then another reference error occurred, so I modified to run with babel build at this time to make `build:atoms` work first than update others.

```
/Users/kazuaki/GitHub/appium-remote-debugger/scripts/common.js:74
export async function seleniumClone () {
^^^^^^

SyntaxError: Unexpected token 'export'
    at Object.compileFunction (node:vm:360:18)
    at wrapSafe (node:internal/modules/cjs/loader:1078:15)
    at Module._compile (node:internal/modules/cjs/loader:1113:27)
    at Module._extensions..js (node:internal/modules/cjs/loader:1203:10)
    at Module.load (node:internal/modules/cjs/loader:1027:32)
    at Module._load (node:internal/modules/cjs/loader:868:12)
    at Module.require (node:internal/modules/cjs/loader:1051:19)
    at require (node:internal/modules/cjs/helpers:103:18)
    at Object.<anonymous> (/Users/kazuaki/GitHub/appium-remote-debugger/scripts/build-atoms.js:1:40)
    at Module._compile (node:internal/modules/cjs/loader:1149:14)
```

scripts/common.js
KazuCocoa: Current await got below error. Should have worked...?

```
/Users/kazuaki/GitHub/appium-remote-debugger/scripts/common.js:174
  })).filter(doesPathMatch);
      ^
TypeError: (intermediate value).filter is not a function
    at atomsCopy (/Users/kazuaki/GitHub/appium-remote-debugger/scripts/common.js:174:7)
    at importAtoms (/Users/kazuaki/GitHub/appium-remote-debugger/scripts/common.js:193:3)
    at /Users/kazuaki/GitHub/appium-remote-debugger/scripts/build-atoms.js:5:3
kazu $ npm run build:atoms
```

mykola-mokhnach: oh, it needs to be wrapped into B.promisify() first before invoking await

package.json
mykola-mokhnach: you are right. Consider replacing usages of `export` with `module.exports = {...}` to avoid babel usage

scripts/common.js
mykola-mokhnach: yes, we could do it in another PR

scripts/common.js
KazuCocoa: updated in [789a7f5](https://github.com/appium/appium-remote-debugger/pull/276/commits/789a7f5a6ea0b53a877fc77037e2f4758b1e95f3)

package.json
KazuCocoa: updated in [789a7f5](https://github.com/appium/appium-remote-debugger/pull/276/commits/789a7f5a6ea0b53a877fc77037e2f4758b1e95f3)

mykola-mokhnach (approved):

mykola-mokhnach:
I assume the PR name prefix should be `fix:` rather than `chore:`

jlipps (approved):
this PR doesn't actually include any changes to the atoms themselves, correct? but fixing the build makes sense!

KazuCocoa:
Correct. Just to make the script works in the current master

PR #283, opened 2023-02-26
chore: add commands for Page

KazuCocoa:
Added all available commands in `Page` from https://github.com/WebKit/WebKit/blob/main/Source/JavaScriptCore/inspector/protocol/Page.json

Current getCookies does not have any input in xcuitest https://github.com/appium/appium-xcuitest-driver/blob/85833718a898c46c3dc2c96769dfd18608bb37bc/lib/commands/cookies.js. Tge Page definition as well, so maybe current remote debgger was wrongly sent null (but it does not matter), maybe.

`overrideUserAgent` probably can be used.

I haven't tested these yet.

Maybe we should move commands related to Page into another file for Page for readability, but this does not include such things.

---

https://github.com/appium/appium-xcuitest-driver/compare/add-some-remotedebugger?expand=1 is an example in the xcuitest driver repository

### DIFF
--- a/lib/remote-debugger.js
+++ b/lib/remote-debugger.js
@@ -208,6 +208,7 @@ class RemoteDebugger extends EventEmitter {
     return this._allowNavigationWithoutReload;
   }
 
+  // Potentially this does not work for mobile safari
   async overrideUserAgent (value) {
     log.debug('Setting overrideUserAgent');
     return await this.rpcClient.send('Page.overrideUserAgent', {
@@ -226,7 +227,7 @@ class RemoteDebugger extends EventEmitter {
   }
 
   async setCookie (cookie) {
-    log.debug('Setting cookies');
+    log.debug('Setting cookie');
     return await this.rpcClient.send('Page.setCookie', {
       appIdKey: this.appIdKey,
       pageIdKey: this.pageIdKey,
--- a/lib/remote-debugger.js
+++ b/lib/remote-debugger.js
@@ -208,6 +208,15 @@ class RemoteDebugger extends EventEmitter {
     return this._allowNavigationWithoutReload;
   }
 
+  async overrideUserAgent (value) {
+    log.debug('Setting overrideUserAgent');
+    return await this.rpcClient.send('Page.overrideUserAgent', {
+      appIdKey: this.appIdKey,
+      pageIdKey: this.pageIdKey,
+      value
+    });
+  }
+
   async getCookies () {
     log.debug('Getting cookies');
     return await this.rpcClient.send('Page.getCookies', {
--- a/lib/remote-debugger.js
+++ b/lib/remote-debugger.js
@@ -225,12 +225,12 @@ class RemoteDebugger extends EventEmitter {
     });
   }
 
-  async setCookies (cookies) {
+  async setCookie (cookie) {
     log.debug('Setting cookies');
     return await this.rpcClient.send('Page.setCookie', {
       appIdKey: this.appIdKey,
       pageIdKey: this.pageIdKey,
-      cookies
+      cookie
     });
   }
 
--- a/lib/remote-debugger.js
+++ b/lib/remote-debugger.js
@@ -208,12 +208,20 @@ class RemoteDebugger extends EventEmitter {
     return this._allowNavigationWithoutReload;
   }
 
-  async getCookies (urls) {
+  async getCookies () {
     log.debug('Getting cookies');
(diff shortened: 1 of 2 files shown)


lib/remote-debugger.js
```diff
@@ -208,12 +208,29 @@ class RemoteDebugger extends EventEmitter {
...
+  async overrideUserAgent (value) {
+    log.debug('Setting overrideUserAgent');
+    return await this.rpcClient.send('Page.overrideUserAgent', {
+      appIdKey: this.appIdKey,
+      pageIdKey: this.pageIdKey,
+      value
+    });
+  }
+
+  async getCookies () {
```
KazuCocoa, line 220: existing getCookies looks like no issue (did not send any URLs param)
```
[HTTP] --> GET /session/107daaaf-b65a-4589-aac6-3981747f2e80/cookie
[HTTP] {}
[debug] [XCUITestDriver@fda2 (107daaaf)] Calling AppiumDriver.getCookies() with args: ["107daaaf-b65a-4589-aac6-3981747f2e80"]
[debug] [XCUITestDriver@fda2 (107daaaf)] Executing command 'getCookies'
[debug] [RemoteDebugger] Getting cookies
[debug] [RemoteDebugger] Sending '_rpc_forwardSocketData:' message to app 'PID:4774', page '1', target 'page-7' (id: 36): 'Page.getCookies'
[debug] [RemoteDebugger] Received data response from send (id: 36): '{"cookies":[]}'
[debug] [RemoteDebugger] Sending to Web Inspector took 10ms
[debug] [XCUITestDriver@fda2 (107daaaf)] Responding to client with driver.getCookies() result: []
[HTTP] <-- GET /session/107daaaf-b65a-4589-aac6-3981747f2e80/cookie 200 14 ms - 12
```

lib/remote-debugger.js
```diff
@@ -208,12 +208,29 @@ class RemoteDebugger extends EventEmitter {
     return this._allowNavigationWithoutReload;
   }
 
-  async getCookies (urls) {
+  async overrideUserAgent (value) {
```
KazuCocoa, line 211: hm, this did not work i guess while the syntax was probably correct (based on the console output). Potentially does not work for mobile safari...?
```
commands.mobileOverrideUserAgent = async function mobileOverrideUserAgent (opts = {}) {
  if (!this.isWebContext()) {
    throw new errors.NotImplementedError();
  }
  const { userAgent } = opts;
  return await this.remote.overrideUserAgent(userAgent);
}
```

KazuCocoa:
Anyway, this implementation itself seems good for now. (I need to investigate setCookies's syntax etc as input later though)

lib/remote-debugger.js
```diff
@@ -208,12 +208,29 @@ class RemoteDebugger extends EventEmitter {
...
+  async getCookies () {
     log.debug('Getting cookies');
     return await this.rpcClient.send('Page.getCookies', {
+      appIdKey: this.appIdKey,
+      pageIdKey: this.pageIdKey
+    });
+  }
+
+  async setCookie (cookie) {
+    log.debug('Setting cookies');
```
mykola-mokhnach, line 229: cookies -> cookie

lib/remote-debugger.js
mykola-mokhnach: have you checked if it is possible to override user agent via Safari Remote debugger?

mykola-mokhnach (approved):

lib/remote-debugger.js
KazuCocoa: I haven't found the way for mobile safari. I saw Safari desktop can change via dev option. Some googled articles explained for mobile safari only accept such change by changing the header info in the Safari.app's info plist. So potentially this feature is not available for mobile safari yet

mykola-mokhnach: we can change the plist though: https://appium.github.io/appium-xcuitest-driver/4.16/execute-methods/#mobile-updatesafaripreferences

Do you know which plist entry it should be?

mykola-mokhnach: > Safari.app's info plist
This one could probably be changed as well, but only on simulator. Also, changing it would modify the agent forever, which is probably not what we want...

KazuCocoa: fyi:
https://www.technipages.com/ios-change-the-user-agent-in-safari is what I found, but hm, my local did not work well for a simulator. (I could mistake something tho)

PR #317, opened 2023-08-24
fix: use correct timeline rpc name

jlipps:
fix appium/appium#15685

did not test this but seems to be the cause of the problem


mykola-mokhnach (approved):

KozhevnikovSergey:
it is amazing. Thank you! )

jlipps:
glad it works!

PR #346, opened 2023-11-14
feat: introduce a wildcard usage for additionalWebviewBundleIds to get webview contexts of other apps

mwakizaka:
- We'd like to switch to a webview context of an app which is launched by `mobile: launchApp` script.
~- We don't like to use `additionalWebViewBundleIds` capability for our case because it's troublesome to get the bundle id of the app before establishing an appium session.~
~- Therefore, I'd like to add a boolean capability named like `includeAllAppsInWebviews`.~
- Therefore, this PR adds a wildcard usage for `additionalWebviewBundleIds` capability.
- Thanks to [isActive check](https://github.com/appium/appium-remote-debugger/blob/f19984e37471949057fbdf4333b816c8132fbc87/lib/mixins/connect.js#L198-L201), I don't think `searchForApp` returns too many `AppPages` 
- If this PR is no problem, 
    - I'd like to send a PR to appium-xcuitest-driver as well to add `includeAllAppsInWebviews` capability.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -173,7 +173,7 @@ function appIdForBundle (bundleId, appDict) {
  */
 function getPossibleDebuggerAppKeys(bundleIds, appDict) {
   if (bundleIds.includes(WILDCARD_BUNDLE_ID)) {
-    log.debug(`Skip checking bundle identifiers because the bundleIds includes a wildcard`);
+    log.debug('Skip checking bundle identifiers because the bundleIds includes a wildcard');
     return _.uniq(Object.keys(appDict));
   }
   let proxiedAppIds = [];
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -167,15 +167,11 @@ function appIdForBundle (bundleId, appDict) {
 /**
  * When includeAllApps is false, it returns appKeys filtered by bundleIds.
  * Otherwise, returns all appKeys in appDict.
- * @param {boolean} includeAllApps
  * @param {string[]} bundleIds
  * @param {Record<string, any>} appDict
  * @returns {string[]}
  */
-function getPossibleDebuggerAppKeys (includeAllApps, bundleIds, appDict) {
-  if (includeAllApps) {
-    return _.uniq(Object.keys(appDict));
-  }
+function getPossibleDebuggerAppKeys(bundleIds, appDict) {
   let proxiedAppIds = [];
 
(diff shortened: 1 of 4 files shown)


KazuCocoa:
Or adding `*` usage instead of adding a new capability?
`additionalWebviewBundleIds: ['*']`, then all bundle ids will be selected

mwakizaka:
Thank you for your comment! Will try adding `*` usage to `additionalWebviewBundleIds` capability.

lib/utils.js
```diff
@@ -164,7 +184,6 @@ function getPossibleDebuggerAppKeys (bundleIds, appDict) {
     WEB_CONTENT_PROCESS_BUNDLE_ID,
     SAFARI_VIEW_PROCESS_BUNDLE_ID,
     SAFARI_VIEW_BUNDLE_ID,
-    WILDCARD_BUNDLE_ID,
```
KazuCocoa, line 167: Can you keep this line? It seems like Safari has `*` case as https://github.com/appium/appium-remote-debugger/pull/179 , not sure all supported iOS versions though.

So only when a user specifies `*` explicitly as the capabilities, the new behavior occurs

mwakizaka: I see, addressed as https://github.com/appium/appium-remote-debugger/pull/346/commits/d92627f80b4bbbefd5bd45df6bff7843d9e8cb0d

test/unit/utils-specs.js
```diff
@@ -57,6 +59,64 @@ describe('utils', function () {
       expect(getDebuggerAppKey('io.appium.bundle', {})).to.not.exist;
     });
   });
+  describe('getPossibleDebuggerAppKeys', function () {
```
KazuCocoa, line 62: 👍

KazuCocoa (commented):
lgtm overall. (waiting for the ci)

mwakizaka:
Confirmed the behaviour as https://gist.github.com/mwakizaka/fa993e9589db3ce0271538e258e9d719

mykola-mokhnach:
there seems to be a test failure in https://github.com/appium/appium-remote-debugger/actions/runs/6902034726/job/18778125000?pr=346

Seems like an issue to me

KazuCocoa:
A retry fixed that. Maybe just flaky?

KazuCocoa (approved):

PR #359, opened 2024-02-14
chore: return timeout error

KazuCocoa:
It did a few more tests.
Then, once this timeout occurs, `driver.title` etc also no longer gets a response. So, maybe it would be helpful to notify if an error occurred explicitly as the response for users so that they can end the session or restart the Safari process rather than silently fail and wait for long response time without any response (and eventually should kill the session or appium process)

### DIFF
--- a/lib/mixins/navigate.js
+++ b/lib/mixins/navigate.js
@@ -213,8 +213,7 @@ async function navToUrl (url) {
         `within ${CONSOLE_ENABLEMENT_TIMEOUT_MS}ms. This might be an Inspector bug.`
       );
       throw new errors.TimeoutError(`Could not enable console events after the page load within ` +
-        `${CONSOLE_ENABLEMENT_TIMEOUT_MS}ms. The Web Inspector/Safari may need to be restarted. ` +
-        `Original error: ${err.message}`);
+        `${CONSOLE_ENABLEMENT_TIMEOUT_MS}ms. The Web Inspector/Safari may need to be restarted.`);
     } else {
       throw err;
     }
--- a/lib/mixins/navigate.js
+++ b/lib/mixins/navigate.js
@@ -4,6 +4,7 @@ import events from './events';
 import { timing, util } from '@appium/support';
 import _ from 'lodash';
 import B from 'bluebird';
+import { errors } from '@appium/base-driver';
 
 const DEFAULT_PAGE_READINESS_TIMEOUT_MS = 20 * 1000;
 const PAGE_READINESS_CHECK_INTERVAL_MS = 50;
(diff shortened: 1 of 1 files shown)


lib/mixins/navigate.js
```diff
@@ -211,6 +212,8 @@ async function navToUrl (url) {
         `Could not enable console events on the page '${this.pageIdKey}' (app '${this.appIdKey}') ` +
         `within ${CONSOLE_ENABLEMENT_TIMEOUT_MS}ms. This might be an Inspector bug.`
       );
+      throw new errors.TimeoutError('Could not enable console event after the page load. ' +
```
mykola-mokhnach, line 215: Lets then bump the default timeout to 15-20 seconds to make sure we don't have any false positives there because of slowly responding device.

mykola-mokhnach: console events

mykola-mokhnach: we should also mention the actual timeout value in the error message

KazuCocoa: makes sense

lib/mixins/navigate.js
```diff
@@ -211,6 +212,9 @@ async function navToUrl (url) {
         `Could not enable console events on the page '${this.pageIdKey}' (app '${this.appIdKey}') ` +
         `within ${CONSOLE_ENABLEMENT_TIMEOUT_MS}ms. This might be an Inspector bug.`
       );
+      throw new errors.TimeoutError(`Could not enable console events after the page load within ` +
+        `${CONSOLE_ENABLEMENT_TIMEOUT_MS}ms. The Web Inspector/Safari may need to be restarted. ` +
+        `Original error: ${err.message}`);
```
mykola-mokhnach, line 217: the original error would look like `timeout` (actually this is the message that we can pass as an argument to the bluebird's `timeout` API), so there is no need to print it here

mykola-mokhnach: Also I am still not 100% convinced we should throw an error if we have actually failed to wait until the page load is completed. Maybe there is some really slow page...
Perhaps in such case it would even make sense to not wait for a response from this RPC call at all...

KazuCocoa: yea, I don't have much slow page like over 10 sec, so I haven't tested well...


At least no response by Web Inspector could occur kind of frequently in Safari 17. When I repeated a URL to repeat navigating 10 times, like this no response case occurred. Based on the Safari's GUI, the progress bar was completed in this initial no response case.

Once this no response case occurred, Safari's GUI no longer show progress bar/load a URL. `driver.title` also had no response so long (at least over a minute, I killed my local Appium server then). So, checking page update with `driver.title` etc does not work well as no response under this condition right now, so I thought raising error was better than keeping no response under case so that making test scenario fail fast

mykola-mokhnach: Let me merge it as is and then I will propose a solution for the "not loaded" case in a different PR

mykola-mokhnach (approved):

lib/mixins/navigate.js
jlipps: isn't there already a user specified page load timeout we could use for page load specifically?

mykola-mokhnach: yes, it is `this.pageLoadMs` and we use it. Check my next change at https://github.com/appium/appium-remote-debugger/pull/360

jlipps: ah got it 👍

PR #369, opened 2024-03-22
fix: Improve page load detection logic

mykola-mokhnach:
I've noticed we always experience page load timeouts for web pages like WDA's /health one. The reason for it is it's actually not a real page, but just a text, so it does not trigger the `Page.loadEventFired` event  like other "normal" pages.

### DIFF
--- a/lib/mixins/navigate.js
+++ b/lib/mixins/navigate.js
@@ -102,14 +102,16 @@ async function waitForDom (startPageLoadTimer) {
 
 /**
  * @this {import('../remote-debugger').RemoteDebugger}
+ * @param {number} [timeoutMs]
  * @returns {Promise<boolean>}
  */
-async function checkPageIsReady () {
+async function checkPageIsReady (timeoutMs) {
   checkParams({appIdKey: this.appIdKey});
 
   const readyCmd = 'document.readyState;';
   try {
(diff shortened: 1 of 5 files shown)


lib/mixins/navigate.js
```diff
@@ -176,6 +181,26 @@ async function navToUrl (url) {
     };
     // https://chromedevtools.github.io/devtools-protocol/tot/Page/#event-loadEventFired
     this.rpcClient?.once('Page.loadEventFired', onPageLoaded);
+    // Pages that have no proper DOM structure do not fire the Page.loadEventFired event
+    // so we rely on the very first event after target change, which is Page.defaultUserPreferencesDidChange
```
KazuCocoa, line 185: Oh... nice
https://github.com/WebKit/WebKit/blob/c125db2d08b6d3ccd41638761132894c54718cc6/Source/JavaScriptCore/inspector/protocol/Page.json#L394

Maye this helps since ios 16.4 https://github.com/search?q=repo%3AWebKit%2FWebKit%20defaultUserPreferencesDidChange&type=code

KazuCocoa (approved):
lg

lib/mixins/navigate.js
mykola-mokhnach: thanks for the hint, I did not know this event is only present since 16.4
I've changed the implementation, so it also works with older system versions

PR #375, opened 2024-04-16
feat: update atoms to selenium 4.19.0 basis

KazuCocoa:
- Update atoms to Selenium 4.19.0 basis

Diffs:

- https://github.com/SeleniumHQ/selenium/pull/12532
- https://github.com/SeleniumHQ/selenium/pull/13211
- https://github.com/SeleniumHQ/selenium/pull/13218
- https://github.com/SeleniumHQ/selenium/pull/13430
- https://github.com/SeleniumHQ/selenium/pull/13792


mykola-mokhnach (approved):
I assume you would want to release this PR together with https://github.com/appium/appium-remote-debugger/pull/376

KazuCocoa:
Sounds good. Let me check Jlipps's thoughts as well in https://github.com/appium/appium-remote-debugger/pull/376
If the react workaround in Appium makes sense, will merge the PR to this branch, and then will merge this to master to publish a new one.
(Agree with the test, bsaically)

KazuCocoa:
I'm going to merge this only since I have closed https://github.com/appium/appium-remote-debugger/pull/376 and moving to https://github.com/appium/appium/issues/20004#issuecomment-2059754083 idea for this case. I think it is better than adding atoms's customization only in Appium Atoms.

jlipps (commented):
Pretty sure this will bring back a bug that I fixed in the atoms, but which the Selenium people did not like and reverted since it caused some other issues. Are all the tests passing with these changes?

KazuCocoa:
Oh, do you mean https://github.com/SeleniumHQ/selenium/pull/12704 ...? Actually it reverted https://github.com/SeleniumHQ/selenium/pull/12557 .

Hm, shall we apply https://github.com/SeleniumHQ/selenium/pull/12557 to our appium atoms? @jlipps 

yea, CI here and https://github.com/appium/appium-xcuitest-driver were fine with this.

KazuCocoa:
ah, maybe `typeof window` .... related errors? Then yea, I removed them to remove diffs after creating atoms on this pr. So actually your fix was applied to this atoms.

Maybe we should note it in this atoms note.

jlipps:
yep exactly. thanks!

PR #376, opened 2024-04-16
feat: react support again

KazuCocoa:
Should we need to keep `React input operation` workaround we had but removed in https://github.com/appium/appium-remote-debugger/pull/268 ...?

So, with this change, as same as we had before, `e.send_keys` to a react element can type all of given chars with one command, although usually it requires sending keys one by one.

So, with this change (and before https://github.com/appium/appium-remote-debugger/pull/268), `e.send_leys '12345'` can type `12345` with one send_keys.
Without this (so current master), it requires `e.send_keys '1';e.send_keys '2';e.send_keys '3';...` one by one to type properly.

Actually in user perspective, this pr may help (and can keep backward compatibility for before https://github.com/appium/appium-remote-debugger/pull/268), but we need to keep this diff as Appium's workaround.

@jlipps any thoughts about this atom? (as https://github.com/appium/appium-remote-debugger/pull/268 's change perspective)


KazuCocoa:
I think https://github.com/appium/appium/issues/20004#issuecomment-2059754083 is more simpler than adding this kind of modification in atoms every time. Closing this PR.

jlipps:
Yes, agree.

PR #395, opened 2024-07-29
fix: Tune page array retrieval

mykola-mokhnach:
It looks like sometimes after we receive a notification about a new app connection no followup pages update notification is delivered. This PR tries to streamline the flow and avoid performing any actions on the app until we actually have the list of pages. 

I've also updated/fixed some typedefs and removed the asynchronous lock as all operations on the `appDict` are synchronous and they don't need any locking

### DIFF
--- a/lib/mixins/connect.js
+++ b/lib/mixins/connect.js
@@ -1,7 +1,6 @@
 import {
   pageArrayFromDict,
   getPossibleDebuggerAppKeys,
-  simpleStringify,
 } from '../utils';
 import events from './events';
 import { timing } from '@appium/support';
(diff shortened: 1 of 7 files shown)


KazuCocoa:
lg for now. I'll do some tests with real devices

lib/mixins/connect.js
```diff
@@ -1,34 +1,19 @@
...
   getPossibleDebuggerAppKeys,
-  simpleStringify,
 } from '../utils';
 import events from './events';
 import { timing } from '@appium/support';
 import { retryInterval, waitForCondition } from 'asyncbox';
 import _ from 'lodash';
 
-
 const APP_CONNECT_TIMEOUT_MS = 0;
```
KazuCocoa, line 10: Can we make this default value to `500` or around? After this turning or with https://github.com/appium/appium-xcuitest-driver/pull/2437, no webview found occurred with `browserName: safari`. It was not always, but `500`+ improved the successful rate on my local device.

KazuCocoa (approved):
Left one comment which improved initial attachment on my local real device.

lg overall.

PR #401, opened 2024-08-04
fix: Properly handle protected accessors from mixin methods

mykola-mokhnach:
It looks like typescript was not really happy about the hack with mixin interfaces (which works for Python, he-he).
So I have implemented them differently using package-private accessors to make the compiler happy


KazuCocoa (approved):
interesting

PR #431, opened 2025-07-12
feat: Add a helper method to verify if javascript execution is blocked

mykola-mokhnach:
This method might be useful to report a better error message in the xcuitest driver


KazuCocoa (approved):
yeah, it's interesting idea

PR #437, opened 2025-08-27
fix: Avoid deadlocks

mykola-mokhnach:
Sending of device commands might be stuck due to various reasons. This fix ensures we never have a deadlock situation in such case

### DIFF
--- a/lib/rpc/rpc-client.js
+++ b/lib/rpc/rpc-client.js
@@ -10,6 +10,7 @@ import AsyncLock from 'async-lock';
 
 const DATA_LOG_LENGTH = {length: 200};
 const MIN_WAIT_FOR_TARGET_TIMEOUT_MS = 30000;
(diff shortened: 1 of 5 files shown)


lib/rpc/rpc-client.js
```diff
@@ -521,57 +524,71 @@ export class RpcClient {
...
+            } else {
+              log.debug(
+                `Provisional target ${targetInfo.targetId}@${appIdKey} is not paused. This might be a problem`
+              );
+            }
           }
         });
+      } catch (e) {
+        log.warn(
+          `Cannot complete the initialization of the provisional target '${targetInfo.targetId}': ${e.message}`
```
KazuCocoa, line 555: Maybe the error could have the timeout info as well (especially in case the `lock.acquire` doesn't print such a timeout value as their error message)

KazuCocoa (approved):
lgtm

PR #498, opened 2026-05-26
chore(remote-debugger): fix compatability with iOS 26

### LINKED ISSUE #497: Inspector.enable / Runtime.enable silently dropped by iOS 26.4.2 over com.apple.webinspector.shim.remote — Target.targetCreated never emitted, selectPage hangs
## Summary

On a USB-connected iPhone running **iOS 26.4.2**, the WebInspector shim path
(`canUseWebInspectorShim` returns `true`, picked because platform >= 18.0)
completes the initial RSD/WebInspector handshake successfully and returns the
correct app dictionary and page listing for a Capacitor WKWebView app. However,
once `selectPage(appIdKey, pageIdKey)` proceeds:

1. The library sends `_rpc_forwardIndicateWebView:` (×2) and `_rpc_forwardSocketSetup:`.
2. iOS responds only with two `_rpc_applicationUpdated:` events (app metadata refresh).
3. **iOS never emits `Target.targetCreated`** for the selected page.
4. `selectPage` therefore waits indefinitely inside `RpcClient.selectPage`'s
   `await new Promise(...)` block, because the `setTimeout`-based fallback in
   `lib/rpc/rpc-client.js` is registered **inside** the
   `ON_PAGE_INITIALIZED_EVENT` handler — it is never armed if no event arrives.

If we force-resolve `selectPage` (by manually emitting `onPageInitialized`)
and then directly send `Inspector.enable` / `Page.enable` / `Runtime.enable`
via `RpcClient.send()` wrapped as `_rpc_forwardSocketData:`, the WebInspector
shim connection stays alive (debug log keeps printing
`No messages received in the last second, continuing to listen...`) but
**iOS 26.4.2 returns no response for any of those Inspector/Runtime/Page
commands sent through `_rpc_forwardSocketData:`** — they are silently dropped.
Consequently `_targets[appIdKey]` is never populated and `execute()` fails with
`No targets found for app 'PID:XXXX'`.

Reads of the iOS WebKit Target protocol suggest that on iOS 26 the
`com.apple.webinspector.shim.remote` channel may now require either:

- a different envelope (not `_rpc_forwardSocketData:`), or
- an explicit `Target.setAutoAttach` / equivalent kick-off before iOS
  begins emitting `Target.targetCreated`, or
- a different RSD service entirely.

Whatever the cause, the current shim implementation in
`appium-remote-debugger@15.10.2` + `appium-ios-remotexpc@2.1.2` cannot
reach a usable `Runtime.evaluate` state on iOS 26.4.2.

I have not yet been able to repro on iOS 18.x / iOS 25.x for direct comparison
(don't have those devices available), but the same code path works on those
versions per other community reports, so this appears to be specific to iOS 26.

---

## Environment

| Component | Version |
| --- | --- |
| `appium-remote-debugger` | `15.10.2` (latest as of 2026-05-25) |
| `appium-ios-remotexpc` | `2.1.2` |
| `@appium/strongbox` | `1.1.1` |
| Node.js | `v22.x` |
| Host OS | macOS Sequoia (Apple Silicon) |
| Device | iPhone (Apple A15), real device over USB |
| Device iOS | **26.4.2** |
| Test app | Capacitor 6 app with `WKWebView.isInspectable = true` (debug build) |
| WebView URL | `capacitor://127.0.0.0/tabs/home` |

Setup that was confirmed working **before** running the repro:

- Developer Mode enabled on the iPhone (Settings → Privacy & Security → Developer Mode).
- Personalized Developer Disk Image mounted (`pymobiledevice3 mounter list`
  shows `IsMounted: true`, image type `Personalized`).
- Trust record paired and current.
- Mac Safari → Develop → \<iPhone\> menu correctly lists the WKWebView with
  title `MOCREO Smart` and URL `capacitor://127.0.0.0/tabs/home`, confirming the
  app's WKWebView is genuinely inspectable end-to-end on iOS 26.4.2 via Safari.
- `appium-ios-remotexpc/scripts/tunnel-creation.mjs --udid <UDID>` running in a
  foreground sudo terminal, registry server on `127.0.0.1:42314`.
- The registry contains the device:
  ```
  curl http://127.0.0.1:42314/remotexpc/tunnels
  { "tunnels": { "<UDID>": { ... rsdPort: 49273 ... } } }
  ```

---

## Steps to Reproduce

1. Start `tunnel-creation.mjs` as root in one terminal, wait for
   `Tunnel Registry Server started on port 42314`.
2. Bootstrap the `tunnelRegistryPort` strongbox item (since when calling
   `appium-remote-debugger` outside of `appium-xcuitest-driver` no one writes
   it). For example via `@appium/strongbox`:
   ```js
   import { BaseItem, strongbox } from '@appium/strongbox';
   await new BaseItem('tunnelRegistryPort', strongbox('appium-xcuitest-driver')).write('42314');
   ```
3. Run the minimal script below.

```js
import { createRemoteDebugger } from 'appium-remote-debugger';

const rd = createRemoteDebugger({
  udid: '<UDID>',
  bundleId: 'com.example.capacitorapp',
  platformVersion: '26',
  isSafari: false,
  includeSafari: false,
  logAllCommunication: true,
  logFullResponse: true,
  targetCreationTimeoutMs: 5000,
}, true);

await rd.connect(15000);
const pages = await rd.selectApp(null, undefined, true);
// pages = [{ id: '4309.1', title: 'MOCREO Smart', url: 'capacitor://127.0.0.0', ... }]
const page = pages.find(p => p.url.startsWith('capacitor://')) ?? pages[0];
const [appIdKey, pageIdKey] = page.id.split('.');

await rd.selectPage(appIdKey, pageIdKey, true);   // ← hangs indefinitely
const dom = await rd.execute('document.title');   // never reached
console.log(dom);
await rd.disconnect();
```

Run with:

```sh
APPIUM_IOS_REMOTEXPC_LOG_LEVEL=debug node repro.mjs
```

---

## Expected Behavior

After `selectPage`, iOS emits `Target.targetCreated` for the selected page.
`RpcClient`'s internal `_targets[appIdKey][pageIdKey]` is populated.
`execute()` sends a `Runtime.evaluate` for the page's target and returns the
JS result.

---

## Actual Behavior

`selectPage` hangs indefinitely (in practice for the full
`targetCreationTimeoutMs * 1.2` window, which is `216000 ms` by default).

Key log slice (debug level enabled, irrelevant lines elided):

```
dbug RemoteDebugger Connecting to WebInspector shim service for device <UDID>
dbug RemoteDebugger Successfully connected to WebInspector shim service
info RemoteDebugger Using WebInspector shim service for iOS 26
dbug RemoteDebugger Sending connection key request
dbug RemoteDebugger Sending '_rpc_reportIdentifier:' message (id: 0): 'setConnectionKey'
dbug WebInspectorService Sending WebInspector message: _rpc_reportIdentifier:
dbug WebInspectorService Connected to WebInspector service
dbug RemoteDebugger Received connected applications list: PID:4309, PID:89
dbug RemoteDebugger Found app id key 'PID:4309' for bundle 'com.example.capacitorapp'
dbug RemoteDebugger Retrieved 1 application within 103ms

dbug RemoteDebugger Sending '_rpc_forwardGetListing:' message to app 'PID:4309' (id: 2): 'connectToApp'
dbug WebInspectorService Sending WebInspector message: _rpc_forwardGetListing:
dbug RemoteDebugger Received response from send (id: 2): '["PID:4309",{"1":{"WIRTitleKey":"MOCREO Smart","WIRTypeKey":"WIRTypeWebPage","WIRURLKey":"capacitor://127.0.0.0","WIRPageIdentifierKey":1,...}}]'
dbug RemoteDebugger Pages changed for PID:4309: undefined -> [{"id":1,...}]

dbug RemoteDebugger Selecting page '1' on app 'PID:4309' and forwarding socket setup
dbug RemoteDebugger Sending '_rpc_forwardIndicateWebView:' (id: 6)
dbug RemoteDebugger Sending '_rpc_forwardIndicateWebView:' (id: 8)
dbug RemoteDebugger Sending '_rpc_forwardSocketSetup:' message to app 'PID:4309', page '1' (id: 10): 'setSenderKey'
dbug WebInspectorService Sending WebInspector message: _rpc_forwardSocketSetup:
dbug RemoteDebugger Waiting up to 216000ms for page '1' of app 'PID:4309' to be selected

dbug RemoteDebugger Notified that an application has been updated     ← only this fires after setSenderKey
dbug RemoteDebugger Notified that an application has been updated
                                                                       ← NO Target.targetCreated event ever
                                                                       ← infinite wait (no event = no setTimeout
                                                                          armed inside onPageInitialized handler)
```

If `selectPage` is force-resolved (by manually emitting `onPageInitialized`
on the rpc client's `_pageSelectionMonitor`), subsequent direct sends of
`Inspector.enable` / `Page.enable` / `Runtime.enable` via
`RpcClient.send(domain, {appIdKey, pageIdKey})` are forwarded as
`_rpc_forwardSocketData:` envelopes, and iOS returns no response for any of
them. The shim's receive loop stays open and prints its keepalive line every
second:

```
dbug RemoteDebugger Sending '_rpc_forwardSocketData:' message to app 'PID:4309', page '1' (id: 12): 'Inspector.enable'
dbug WebInspectorService Sending WebInspector message: _rpc_forwardSocketData:
dbug WebInspectorService No messages received in the last second, continuing to listen...
dbug WebInspectorService No messages received in the last second, continuing to listen...
dbug WebInspectorService No messages received in the last second, continuing to listen...
(continues forever)
```

`execute()` therefore fails (or hangs in `waitForPage`) with
`No targets found for app 'PID:4309'`.

---

## Diagnostic Evidence Summary

1. **Lower layers are fine.** RSD tunnel, RemoteXPC connection, service
   discovery via `findService(WebInspectorService.RSD_SERVICE_NAME)`,
   `_rpc_reportIdentifier:` handshake, `_rpc_forwardGetListing:` for the app
   list — **all of these get correct responses from iOS 26.4.2**. The shim
   socket stays open and `WebInspectorService` keeps polling without error.

2. **Failure mode is specific to "control-plane" commands going through
   `_rpc_forwardSocketData:`.** Any Inspector/Page/Runtime command sent in
   that envelope receives no response and no asynchronous event.

3. **No `Target.targetCreated` event arrives** after `_rpc_forwardSocketSetup:`
   succeeds. On iOS 18.x as documented elsewhere in this codebase, iOS auto-
   emits this event after setSenderKey, which is what unblocks
   `onTargetCreated` → `_initializePage` → emits `ON_PAGE_INITIALIZED_EVENT`.

4. **Safari's own Web Inspector still works** against the same device + app
   end-to-end, so the WebView itself is inspectable on iOS 26.4.2; whatever
   protocol Safari uses, this library no longer matches it on this iOS
   version.

5. The `selectPage` event-wait construction in
   `lib/rpc/rpc-client.js` (line ~660 in 15.10.2) has the `setTimeout`
   for "Page has not been selected within Xms. Continuing anyway" placed
   **inside** the `onPageInitialized` event handler, so when no event is ever
   received the timeout never fires. This is independent of the iOS 26 issue
   above but means that on environments where the protocol simply fails
   silently, the call hangs for the full `targetCreationTimeoutMs * 1.2`
   (default 3.6 minutes) with no diagnostic surfaced. Even setting
   `targetCreationTimeoutMs: 5000` does not help because the timeout setup
   path requires at least one event to be received first. Might be worth a
   separate fix to arm the timeout unconditionally at the start of the wait.

---

## Workarounds I'm aware of

None that produce a working `execute()` on iOS 26.4.2. Options that were tried
and did not resolve the underlying issue:

- Falling back to the legacy `RpcClientRealDevice` (by passing empty
  `platformVersion`) — the legacy `com.apple.webinspector` lockdownd service
  is gone on iOS 18+, so this just errors out earlier.
- Substituting `ios-webkit-debug-proxy 1.9.2` for the entire transport — that
  proxy has no translation table entry for iOS 17+ `Target.sendMessageToTarget`
  multiplexing and silently `ok`-acks the outer command without forwarding the
  inner payload. Different layer, same outcome.
- Manually emitting `onPageInitialized` to bypass the hang — proceeds, but
  `_targets` is empty so `execute()` fails immediately as quoted above.
- Manually priming `Inspector.enable` etc. via `rpcClient.send()` after
  `selectPage` returns — iOS does not respond, see log above.

---

## Possible Root Cause Hypotheses

I'm not sure which is right; offering these in case it's helpful triage signal:

1. **iOS 26 changed the wrapper format for inspector messages on
   `webinspector.shim.remote`** — `_rpc_forwardSocketData:` no longer the
   correct envelope, or now requires an additional WIR\* key.
2. **iOS 26 now expects an explicit `Target.setAutoAttach` (or similar
   protocol-level kickoff) before emitting `Target.targetCreated`** — i.e. the
   automatic emission after setSenderKey is gone, and replaced by an opt-in.
3. **A different RSD service is now used** for live page inspection — perhaps
   the `webinspector.shim.remote` name no longer maps to the active page
   inspector, only to the legacy app-list path (which is still working here).

Capturing a Safari → iPhone Wireshark/`dump-rsd-traffic` session against iOS
26.4.2 would probably disambiguate quickly. I don't have such a capture
available right now but happy to provide one if it would unblock the fix.

---

## Related

- `appium-remote-debugger` v15.6.0 introduced the shim path with
  `canUseWebInspectorShim(>=18.0)`.
- Internal user-side write-up (mine, has Chinese text, English summary at top):
  see ENG-003 in our team's engineering notes. Can extract verbatim on request.

---

## Secondary Issue (full Appium stack also blocked on iOS 26, by a different layer)

I followed up by trying the issue through the full **Appium 3.4.2 + appium-xcuitest-driver 11.7.1 + WDA 13.x + Xcode 26** stack on the same device, to rule out
that the bug was specific to using `appium-remote-debugger` standalone. The full
stack reaches an **earlier** failure point and never gets to the inspector layer:
`xcodebuild` cannot build/sign WDA 13.x against this iOS 26.4.2 device.

### Versions

| Component | Version |
| --- | --- |
| `appium` | 3.4.2 |
| `appium-xcuitest-driver` | 11.7.1 |
| `appium-webdriveragent` (nested via xcuitest) | 13.x |
| Xcode | 26 (SDK `iphoneos26.5`, confirmed in xcodebuild's `sdk_canonicalName`) |
| Node.js | 24.8.0 |
| iPhone | iPhone14,7 on iOS 26.4.2 (build 23E261) |

### Repro

Minimal POC: `scripts/poc-ios-webview-via-appium.mjs` in the same repo (W3C HTTP
via raw `fetch`, no webdriverio dep). Caps:

```js
{
  platformName: 'iOS',
  'appium:automationName': 'XCUITest',
  'appium:udid': '<UDID>',
  'appium:platformVersion': '26',
  'appium:bundleId': 'com.example.capacitorapp',
  'appium:useNewWDA': false,
  'appium:showXcodeLog': true,
  'appium:xcodeOrgId': '<10-char team id from `security find-identity`>',
  'appium:xcodeSigningId': 'Apple Development',
}
```

`xcodeOrgId` is sourced from the real signing cert on this Mac:

```
$ security find-identity -v -p codesigning | grep "Apple Development"
1) ... "Apple Development: foo@example.com (XCS73H34SD)"
```

### Failure log (relevant lines from `appium:showXcodeLog`)

```
[Xcode] note: Using codesigning identity override: Apple Development
[Xcode] /Users/.../appium-webdriveragent/WebDriverAgent.xcodeproj: error:
        No profiles for 'com.facebook.WebDriverAgentRunner.xctrunner' were found:
        Xcode couldn't find any iOS App Development provisioning profiles matching
        'com.facebook.WebDriverAgentRunner.xctrunner'. Automatic signing is disabled
        and unable to generate a profile. To enable automatic signing, pass
        -allowProvisioningUpdates to xcodebuild.
        (in target 'WebDriverAgentRunner' from project 'WebDriverAgent')

[Xcode] /Users/.../appium-webdriveragent/WebDriverAgent.xcodeproj: error:
        No signing certificate "iOS Development" found: No "iOS Development"
        signing certificate matching team ID "<TEAM_ID>" with a private key was found.
        (in target 'WebDriverAgentLib' from project 'WebDriverAgent')

[Xcode] ** TEST BUILD FAILED **
[Xcode] xcodebuild exited with code '65'
```

Two distinct failures:

1. **`Automatic signing is disabled and unable to generate a profile`** —
   `xcuitest-driver` does not pass `-allowProvisioningUpdates` to `xcodebuild`,
   so Xcode 26's stricter auto-provisioning gate refuses to generate a profile
   for `com.facebook.WebDriverAgentRunner.xctrunner` even though a valid team
   is provided via `xcodeOrgId`.
2. **`No signing certificate "iOS Development" found ... in target 'WebDriverAgentLib'`** —
   the `WebDriverAgentLib` target in
   `appium-webdriveragent/WebDriverAgent.xcodeproj` still references the
   long-deprecated **"iOS Development"** identity name. Apple renamed this to
   **"Apple Development"** years ago. The command-line override
   `CODE_SIGN_IDENTITY=Apple Development` xcuitest correctly emits (visible in
   the log) does not override this target-level setting on Xcode 26.

These look like two independent shortfalls of the Appium stack against Xcode 26:

- `appium-xcuitest-driver` / `appium-webdriveragent` should pass
  `-allowProvisioningUpdates` to xcodebuild by default (or expose a cap that
  defaults to true) so Xcode 26 can auto-generate the WDA profile.
- The `appium-webdriveragent` `WebDriverAgent.xcodeproj` should be migrated from
  `"iOS Development"` to `"Apple Development"` (or `"$(CODE_SIGN_IDENTITY)"`
  variable-form) so the command-line override actually applies.

Both are independent of the primary `Inspector.enable` silently-dropped issue
above. Combined, they form the current state of iOS 26 support in this ecosystem:
even before reaching the inspector layer that the primary issue describes, the
WDA real-device build is blocked.

Happy to file these as separate issues against
`appium-xcuitest-driver` and `appium-webdriveragent` if that's more appropriate.

Delta456:
Fixes https://github.com/appium/appium-remote-debugger/issues/497

iOS 26 introduced changes which caused the remote debugger to hang on forever.

The fix was to enforce integer conversions for `WIRPageIdentifierKey` and prevent the infinite hang by passing false to `Target.setPauseOnStart` as the behavior is changed for iOS 26.

I also added a test from the issue so that it can tested.

### DIFF
--- a/lib/rpc/remote-messages.ts
+++ b/lib/rpc/remote-messages.ts
@@ -97,8 +97,8 @@ export class RemoteMessages {
         WIRApplicationIdentifierKey: appIdKey,
         WIRConnectionIdentifierKey: connId,
         WIRSenderKey: senderId,
-        WIRPageIdentifierKey: pageIdKey,
-        WIRAutomaticallyPause: false,
+        WIRPageIdentifierKey: pageIdKey != null ? Number(pageIdKey) : pageIdKey,
+        WIRMessageDataTypeChunkSupportedKey: 0,
       },
       __selector: '_rpc_forwardSocketSetup:',
     };
@@ -124,7 +124,7 @@ export class RemoteMessages {
         WIRApplicationIdentifierKey: appIdKey,
         WIRIndicateEnabledKey: enabled == null ? true : enabled,
         WIRConnectionIdentifierKey: connId,
-        WIRPageIdentifierKey: pageIdKey,
+        WIRPageIdentifierKey: pageIdKey != null ? Number(pageIdKey) : pageIdKey,
       },
       __selector: '_rpc_forwardIndicateWebView:',
     };
@@ -192,7 +192,7 @@ export class RemoteMessages {
         WIRConnectionIdentifierKey: connId,
         WIRSenderKey: senderId,
         WIRApplicationIdentifierKey: appIdKey,
-        WIRPageIdentifierKey: pageIdKey,
+        WIRPageIdentifierKey: pageIdKey != null ? Number(pageIdKey) : pageIdKey,
       },
       __selector: '_rpc_forwardSocketData:',
     };
@@ -228,7 +228,7 @@ export class RemoteMessages {
         WIRConnectionIdentifierKey: connId,
         WIRSenderKey: senderId,
         WIRApplicationIdentifierKey: appIdKey,
-        WIRPageIdentifierKey: pageIdKey,
+        WIRPageIdentifierKey: pageIdKey != null ? Number(pageIdKey) : pageIdKey,
       },
       __selector: '_rpc_forwardSocketData:',
     };
@@ -255,7 +255,7 @@ export class RemoteMessages {
         WIRConnectionIdentifierKey: connId,
         WIRSenderKey: senderId,
         WIRApplicationIdentifierKey: appIdKey,
-        WIRPageIdentifierKey: pageIdKey,
+        WIRPageIdentifierKey: pageIdKey != null ? Number(pageIdKey) : pageIdKey,
       },
       __selector: '_rpc_forwardSocketData:',
     };
--- a/lib/rpc/remote-messages.ts
+++ b/lib/rpc/remote-messages.ts
@@ -20,6 +20,13 @@ function omitNilValues<T extends Record<string, any>>(obj: T): Partial<T> {
   ) as Partial<T>;
 }
 
+function toPageIdKey(pageIdKey?: PageIdKey): {WIRPageIdentifierKey: number} | Record<never, never> {
+  if (pageIdKey == null || pageIdKey === '' || Number.isNaN(+pageIdKey)) {
+    return {};
+  }
+  return {WIRPageIdentifierKey: +pageIdKey};
+}
+
 // mapping of commands to the function for getting the command
 // defaults to `getMinimalCommand`, so no need to have those listed here
 const COMMANDS = {
@@ -97,7 +104,7 @@ export class RemoteMessages {
         WIRApplicationIdentifierKey: appIdKey,
         WIRConnectionIdentifierKey: connId,
         WIRSenderKey: senderId,
-        WIRPageIdentifierKey: pageIdKey != null ? Number(pageIdKey) : pageIdKey,
+        ...toPageIdKey(pageIdKey),
         WIRMessageDataTypeChunkSupportedKey: 0,
       },
       __selector: '_rpc_forwardSocketSetup:',
@@ -124,7 +131,7 @@ export class RemoteMessages {
         WIRApplicationIdentifierKey: appIdKey,
         WIRIndicateEnabledKey: enabled == null ? true : enabled,
         WIRConnectionIdentifierKey: connId,
-        WIRPageIdentifierKey: pageIdKey != null ? Number(pageIdKey) : pageIdKey,
+        ...toPageIdKey(pageIdKey),
       },
       __selector: '_rpc_forwardIndicateWebView:',
     };
@@ -192,7 +199,7 @@ export class RemoteMessages {
         WIRConnectionIdentifierKey: connId,
         WIRSenderKey: senderId,
         WIRApplicationIdentifierKey: appIdKey,
-        WIRPageIdentifierKey: pageIdKey != null ? Number(pageIdKey) : pageIdKey,
+        ...toPageIdKey(pageIdKey),
       },
       __selector: '_rpc_forwardSocketData:',
     };
@@ -228,7 +235,7 @@ export class RemoteMessages {
         WIRConnectionIdentifierKey: connId,
         WIRSenderKey: senderId,
         WIRApplicationIdentifierKey: appIdKey,
-        WIRPageIdentifierKey: pageIdKey != null ? Number(pageIdKey) : pageIdKey,
+        ...toPageIdKey(pageIdKey),
       },
       __selector: '_rpc_forwardSocketData:',
     };
@@ -255,7 +262,7 @@ export class RemoteMessages {
         WIRConnectionIdentifierKey: connId,
         WIRSenderKey: senderId,
         WIRApplicationIdentifierKey: appIdKey,
-        WIRPageIdentifierKey: pageIdKey != null ? Number(pageIdKey) : pageIdKey,
+        ...toPageIdKey(pageIdKey),
       },
       __selector: '_rpc_forwardSocketData:',
     };
--- a/lib/rpc/rpc-client.ts
+++ b/lib/rpc/rpc-client.ts
@@ -647,11 +647,15 @@ export class RpcClient {
     this.targets[appIdKey][pageIdKey] = targetInfo.targetId;
 
     try {
-      await this.send('Target.setPauseOnStart', {
-        pauseOnStart: true,
-        appIdKey,
-        pageIdKey,
-      });
+      await this.send(
+        'Target.setPauseOnStart',
+        {
+          pauseOnStart: true,
+          appIdKey,
+          pageIdKey,
+        },
+        false,
+      );
     } catch (e: any) {
       log.debug(
         `Cannot setup pause on start for app '${appIdKey}' and page '${pageIdKey}': ${e.message}`,
--- a/test/integrated/real-device-safari-specs.ts
+++ b/test/integrated/real-device-safari-specs.ts
@@ -7,6 +7,12 @@ const PLATFORM_VERSION = process.env.PLATFORM_VERSION ?? '26';
 describe('Real device Safari remote debugger', function () {
   this.timeout('2m');
 
+  before(function () {
+    if (!process.env.DEVICE_UDID) {
+      this.skip();
+    }
+  });
+
   it('should connect to Safari and execute JavaScript', async function () {
     const rd = createRemoteDebugger(
       {
--- a/test/functional/real-device-safari-specs.ts
+++ b/test/functional/real-device-safari-specs.ts
@@ -1,0 +1,41 @@
+import {createRemoteDebugger} from '../../lib';
+import {expect} from 'chai';
+
+const UDID = process.env.DEVICE_UDID ?? '';
+const PLATFORM_VERSION = process.env.PLATFORM_VERSION ?? '26';
+
+describe('Real device Safari remote debugger', function () {
+  this.timeout('2m');
+
+  it('should connect to Safari and execute JavaScript', async function () {
+    const rd = createRemoteDebugger(
+      {
+        udid: UDID,
+        bundleId: 'com.apple.mobilesafari',
+        platformVersion: PLATFORM_VERSION,
+        isSafari: true,
+        includeSafari: false,
+        logAllCommunication: false,
+        logFullResponse: false,
+        targetCreationTimeoutMs: 5000,
+      },
+      true,
+    );
+
+    try {
+      await rd.connect(15000);
+
+      const pages = await rd.selectApp(null, undefined, true);
+      expect(pages).to.have.length.above(0, 'No Safari pages found');
+
+      const page = pages[0];
+      const [appIdKey, pageIdKey] = String(page.id).split('.');
+      await rd.selectPage(appIdKey, pageIdKey, true);
+
+      const title = await rd.execute('document.title');
+      expect(title).to.be.a('string').and.not.be.empty;
+    } finally {
+      await rd.disconnect();
+    }
+  });
+});


mykola-mokhnach:
It looks like no workflows were executed on this PR due to GH outage. @Delta456 Could you please add some fixes to it to reinit the CI?

lib/rpc/remote-messages.ts
```diff
@@ -97,8 +97,8 @@ export class RemoteMessages {
         WIRApplicationIdentifierKey: appIdKey,
         WIRConnectionIdentifierKey: connId,
         WIRSenderKey: senderId,
-        WIRPageIdentifierKey: pageIdKey,
-        WIRAutomaticallyPause: false,
+        WIRPageIdentifierKey: pageIdKey != null ? Number(pageIdKey) : pageIdKey,
```
mykola-mokhnach: does it even make sense to include WIRPageIdentifierKey if it is set to an invalid number/non number?

mykola-mokhnach: perhaps something like

```
...(pageIdKey == null || pageIdKey === '' || Number.isNaN(+pageIdKey) ? {} : {WIRPageIdentifierKey: +pageIdKey}),
```

would make more sense

lib/rpc/remote-messages.ts
```diff
@@ -97,8 +97,8 @@ export class RemoteMessages {
...
+        WIRMessageDataTypeChunkSupportedKey: 0,
```
mykola-mokhnach, line 108: is it enough to only add this property here or it should also be added to records below?

lib/rpc/remote-messages.ts
```diff
@@ -124,7 +124,7 @@ export class RemoteMessages {
         WIRApplicationIdentifierKey: appIdKey,
         WIRIndicateEnabledKey: enabled == null ? true : enabled,
         WIRConnectionIdentifierKey: connId,
-        WIRPageIdentifierKey: pageIdKey,
+        WIRPageIdentifierKey: pageIdKey != null ? Number(pageIdKey) : pageIdKey,
```
mykola-mokhnach: maybe it makes sense to have this condition moved to a shared helper function

lib/rpc/rpc-client.ts
```diff
@@ -647,11 +647,15 @@ export class RpcClient {
...
-        pageIdKey,
-      });
+      await this.send(
+        'Target.setPauseOnStart',
+        {
+          pauseOnStart: true,
+          appIdKey,
+          pageIdKey,
+        },
+        false,
```
mykola-mokhnach, line 657: is this change necessary?

test/integrated/real-device-safari-specs.ts
```diff
@@ -0,0 +1,41 @@
+import {createRemoteDebugger} from '../../lib';
+import {expect} from 'chai';
+
+const UDID = process.env.DEVICE_UDID ?? '';
+const PLATFORM_VERSION = process.env.PLATFORM_VERSION ?? '26';
+
+describe('Real device Safari remote debugger', function () {
+  this.timeout('2m');
+
+  it('should connect to Safari and execute JavaScript', async function () {
```
mykola-mokhnach, line 16: Please isolate this test, so it does not execute as part of standard CI workflow since the latter can only run on simulators

lib/rpc/rpc-client.ts
Delta456: Yes, it's required. iOS 26 no longer sends any response for this command else it will hang indefinitely.

lib/rpc/remote-messages.ts
Delta456: Yes, it is enough only to add this property because it is a part of `_rpc_forwardSocketSetup` which is this only.

lib/rpc/remote-messages.ts
Delta456: Good idea

test/integrated/real-device-safari-specs.ts
Delta456: Making an integrated tests folder for this

lib/rpc/remote-messages.ts
Delta456: Ok

mykola-mokhnach (approved):

PR #511, opened 2026-07-10
chore: Drop mocha

mykola-mokhnach:


package.json
```diff
@@ -53,8 +53,8 @@
     "format": "prettier -w ./lib ./test",
     "format:check": "prettier --check ./lib ./test",
     "prepare": "npm run build",
-    "test": "mocha --exit --timeout 1m \"./test/unit/**/*-specs.ts\"",
-    "e2e-test": "mocha --exit --timeout 1m \"./test/functional/**/*-specs.ts\"",
+    "test": "node --test --enable-source-maps --test-force-exit --test-timeout=60000 \"./build/test/unit/**/*.spec.js\"",
+    "e2e-test": "node --test --enable-source-maps --test-force-exit --test-concurrency=1 --test-timeout=300000 \"./build/test/functional/**/*.spec.js\"",
```
eglitise: The timeout here is bumped from 1 minute to 5 minutes, is that okay?

mykola-mokhnach: I think it should be much higher. Unlike in mocha where timeout is set per test this timeout is set per whole suite, and e2e tests on simulator are slow

KazuCocoa (approved):

PR #522, opened 2026-08-04
feat: Vendor Selenium atoms

mykola-mokhnach:
## Summary

- Replace the Bazel-based atom build (`scripts/build-selenium.mjs` clones SeleniumHQ/selenium, `scripts/build-atoms.mjs` shells out to `bazel`) with a self-contained, pure Node.js pipeline. Upstream Selenium's JS atoms are effectively unmaintained, so this package no longer syncs from it — `atoms/src/` is now a one-time, hand-pruned vendored snapshot (~120 files, computed via Closure Compiler's own `--dependency_mode=PRUNE`/`--output_manifest` so only what's actually reachable from the 44 atoms we ship is included, not Selenium's full `javascript/` tree or the full Closure Library) that this package owns and maintains directly going forward.
- New `scripts/build-atoms.mjs` compiles each atom from `atoms/src/` with `google-closure-compiler` (native binary via `optionalDependencies`, no Java/Bazel required) using `ADVANCED_OPTIMIZATIONS` + the mobile-Safari-only `ASSUME_MOBILE_WEBKIT` define, reproducing Selenium's `closure_fragment` wrapper. Compiled output still lands in `atoms/*.js`, unchanged from `lib/atoms.ts`'s perspective. Rebuilds are deterministic and compile in parallel (bounded to CPU count via `asyncbox`'s `asyncmap`), cutting `npm run build:atoms` from ~47s to ~6s.
- Add `test/unit/atoms.spec.ts`: 27 jsdom-backed green-path tests exercising all 44 compiled atoms through the real `getScriptForAtom`/`convertJavascriptEvaluationResult` production path (locators, element state, interaction, frames, storage, script execution, HTML5 storage/appcache/SQL/geolocation, element cache). Runs on every push/PR via `npm test`, no simulator needed.
- Expand `test/functional/safari-e2e.spec.ts` with 5 more real-Safari-in-Simulator cases (visibility/enabled state, checkbox click, clear, local/session storage round-trip, active element/default content).
- CI: delete the manual `update-atoms.yml` Selenium-sync workflow; add a `verify-atoms` job to `unit-test.yml`, path-filtered to only run when `atoms/**` or `scripts/build-atoms.mjs` changes — rebuilds and fails on drift between `atoms/src/` and the committed `atoms/*.js`, no-op otherwise.
- Rewrite `docs/update-atoms.md`, update `README.md`/`development-notes.md` to describe the new self-owned model; drop `atoms/lastupdate` (no longer meaningful without a live upstream sync — provenance is now just git history).


atoms/src/third_party/closure/goog/disposable/idisposable.js
```diff
@@ -0,0 +1,45 @@
+/**
+ * @license
+ * Copyright The Closure Library Authors.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview Definition of the disposable interface.  A disposable object
+ * has a dispose method to to clean up references and resources.
+ */
```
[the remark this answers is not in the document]
mykola-mokhnach: Fixed in 74fc1d4 — removed the duplicated "to to".

atoms/src/webdriver/atoms/storage/session_storage.js
```diff
@@ -0,0 +1,88 @@
...
+ *     otherwise null.
+ */
+webdriver.atoms.storage.session.getItem = function (key) {
+  return bot.storage.getSessionStorage().getItem(key);
+};
+
+/**
+ * Returns an array of keys of all keys of the sessionStorage object.
+ * @return {Array.<string>} The array of stored keys..
+ */
```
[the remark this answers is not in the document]
mykola-mokhnach: Fixed in 74fc1d4 — removed the stray extra period.

atoms/src/atoms/html5/database.js
```diff
@@ -0,0 +1,139 @@
...
+ *     default value is 5MB.
+ * @param {!Window=} opt_window The window associated with the database;
+ *     defaults to the main window.
+ * @return {!Database} The object to access the web database.
+ *
+ */
+bot.storage.database.openOrCreate = function (databaseName, opt_version, opt_displayName, opt_size, opt_window) {
+  var version = opt_version || '';
+  var displayName = opt_displayName || databaseName + 'name';
+  var size = opt_size || 5 * 1024 * 1024;
```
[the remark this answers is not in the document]
mykola-mokhnach: Good catch — fixed the doc in 74fc1d4 to describe the actual (upstream, unchanged) behavior: `opt_displayName` defaults to `databaseName + 'name'`, not `databaseName`. Left the implementation as-is since it's vendored atom behavior (`execute_sql`) I don't want to silently change.

.github/workflows/unit-test.yml
```diff
@@ -6,6 +6,38 @@ on:
   workflow_dispatch:
 
 jobs:
+  verify-atoms:
```
eglitise: Would it not be cleaner to move this to its own workflow? It's not quite the same as other unit tests

eglitise: If this is applied, the last section in `docs/update-atoms.md` should also be updated accordingly

eglitise: It's also worth taking into account Copilot's suppressed comment:
> The verify-atoms job is skipped unless atoms/** or scripts/build-atoms.mjs changes, but changes to package.json (e.g., bumping google-closure-compiler) can also change the generated atoms/*.js output. Consider including package.json in the path filter so CI catches drift when the compiler/tooling version changes.

mykola-mokhnach: Split into its own workflow in 7764284 — see `verify-atoms.yml`.

mykola-mokhnach: Updated `docs/update-atoms.md`'s CI section in 7764284 to reference `verify-atoms.yml`.

mykola-mokhnach: Added `package.json` to the path filter in 7764284.

.github/workflows/verify-atoms.yml
```diff
@@ -0,0 +1,93 @@
...
+    steps:
+    - uses: actions/checkout@v7
+    - uses: dorny/paths-filter@v3
+      id: filter
+      with:
+        filters: |
+          atoms:
+          - 'atoms/**'
+          - 'scripts/build-atoms.mjs'
+          - 'package.json'
```
eglitise: It'd be much simpler to just move these paths to the `on.push.paths`/`on.pull_request.paths` fields. That way the job won't be triggered at all if the atoms aren't changed

mykola-mokhnach: Good call — switched to `on.push.paths`/`on.pull_request.paths` in 0700073, dropping the `changes` job and per-job `if` conditions entirely.

PR #525, opened 2026-08-07
feat: Make atoms return W3C-compatible element and error responses

mykola-mokhnach:
## Summary

  Makes this package's Selenium atoms return W3C-compatible responses (elements and errors) instead of legacy JSONWP-only shapes, in support of eventually dropping legacy JSONWP protocol support from `@appium/base-driver`. Also fixes an unrelated crash bug found while verifying the change end-to-end against a real Appium server and `appium-xcuitest-driver`.

  ## Why

  `@appium/base-driver`'s JSONWP support is legacy and slated for eventual removal. This package's vendored atoms (`atoms/src/`) were still 100% legacy-shaped: numeric `bot.ErrorCode` statuses in a `{status, value}` envelope, and elements as `{ELEMENT: id}` only — with no W3C `element-6066-11e4-a52e-4f735466cecf` key anywhere. The TS consumer layer (`lib/utils/javascript.ts`) constructed errors via `errorFromMJSONWPStatusCode`, an export `@appium/base-driver` explicitly labels as an "MJSONWP export" and a candidate for removal.

  ## What changed

  **Atoms (`atoms/src/`, rebuilt into `atoms/*.js`):**
  - `bot.inject.wrapValue` now dual-keys element results with both the legacy `ELEMENT` key and the W3C `element-6066-11e4-a52e-4f735466cecf` key.
  - `bot.inject.unwrapValue` and `find_element.js`'s root-element handling now accept *either* key as input.
  - `bot.inject.wrapError` now includes a W3C error signature string (e.g. `'no such element'`) alongside the existing legacy numeric `status`, sourced from the already-vendored (previously unused) `bot.Error.CODE_TO_STATE_` map via a new `bot.Error.stateForCode` helper.

  **TS consumer (`lib/utils/javascript.ts`):**
  - `convertJavascriptEvaluationResult` now constructs errors via `errorFromW3CJsonCode` instead of the MJSONWP-labeled `errorFromMJSONWPStatusCode`, decoupling this package from base-driver's legacy-specific exports ahead of their removal.
  - The failure trigger is `status !== 0` (unambiguous, still present on every atom response today) — `value.error` is only consulted to pick the right error class/message, never as an independent trigger. This avoids a false-positive class of bug where a script's own legitimate return value (e.g. `{error: "some app data"}`) could otherwise be mistaken for a WebDriver failure.

  **Bug fix (`lib/mixins/navigate.ts`):**
  - Found while verifying the above against a real Appium server: a fire-and-forget `Runtime.evaluate` call in `navToUrl` had no `.catch()`, so a rejection (e.g. the RPC connection dropping mid-navigation, which happens routinely around window/tab switches) became an **unhandled promise rejection that crashed the entire Appium server process**. Fixed by attaching a `.catch()` that logs and swallows the failure — navigation readiness is already tracked separately via the `Page.loadEventFired` listener and timeout, so this doesn't change intended behavior, just stops it from taking the whole server down.

  ## Backward compatibility

  No breaking changes. Confirmed via:
  - `appium-xcuitest-driver`'s element handling (`getElementId`/`hasElementId`/`util.wrapElement`/`unwrapElement`) already reads both key shapes and only ever sends the legacy `{ELEMENT: id}` shape 
No breaking changes. Confirmed via:
- `appium-xcuitest-driver`'s element handling (`getElementId`/`hasElementId`/`util.wrapElement`/`unwrapElement`) already reads both key shapes and only ever sends the legacy `{ELEMENT: id}` shape into atoms — both keep working unmodified.
- Error handling in `appium-xcuitest-driver` only ever catches typed Appium error classes (`errors.NoSuchElementError`, etc.), never inspects raw status codes — unaffected by the internal error-construction change.
- No `@appium/base-driver` or `appium-xcuitest-driver` code changes were needed.


eglitise (approved):
Note: it's worth synchronizing `bot.Error.State` with the current W3C protocol in a follow-up PR (they were last synchronized in 2015)

atoms/src/atoms/inject.js
```diff
@@ -68,6 +68,14 @@ bot.inject.JsonWindow;
  */
 bot.inject.ELEMENT_KEY = 'ELEMENT';
 
+/**
+ * Key used to identify DOM elements in the W3C WebDriver wire protocol.
+ * @type {string}
+ * @const
+ * @see https://w3c.github.io/webdriver/webdriver-spec.html#elements
+ */
+bot.inject.W3C_ELEMENT_KEY = 'element-6066-11e4-a52e-4f735466cecf';
```
KazuCocoa, line 77: oh, i see. here will add the key

PR #536, opened 2026-08-12
chore: exercise atoms against Angular-rendered DOM

mykola-mokhnach:
Mirrors #535's React fixture with a zoneless standalone Angular component (text/checkbox/select/radio/OTP-autofocus/disabled/conditional-style/submit, plus @if/@for control-flow blocks that add/remove real DOM nodes). Mounting uses provideZonelessChangeDetection since zone.js's global patching fights Node's test runner; tests flush change detection explicitly via tick().

Exercising clear() against it surfaced a real gap: it only fired a 'change' event, never 'input', so frameworks that bind to 'input' for live updates (Angular, Vue, vanilla — React is the outlier, also treating 'change' as an input-changed signal) never saw a cleared field. Fixed to fire input events too, matching how a real user's select-all + delete behaves.


atoms/src/core/action.ts
```diff
@@ -35,6 +35,12 @@ export function clear(element: Element): void {
   if (el.value) {
     legacyFocusOnElement(element);
     setElementValue(element, '');
+    // A real user clearing a field (e.g. select-all + delete) fires 'input' as the value changes,
+    // then 'change' once the field loses focus. Frameworks bound to 'input' for live updates (most
+    // of them — React is unusual in also treating 'change' as an input-changed signal) never see
+    // the clear without it.
+    fire(element, EventType.TEXTINPUT);
```
KazuCocoa: Could we omit the `textInput` event here and keep only `input`?

Clearing via select-all + Delete is a deletion operation, while the UI Events algorithm only emits the obsolete `textInput` event for text insertion—not deletion. Dispatching it here could trigger legacy insertion handlers in addition to the correct `input` handler.

It would also be useful to add an event-sequence assertion covering this behavior.

Reference: https://w3c.github.io/uievents/event-algo.html#fire-key-input-events

--

I got comment the above by Codex. I'd like to double check if this is vald

mykola-mokhnach: Good catch, and confirmed valid. This codebase already draws exactly that insertion-vs-deletion line elsewhere: `updateOnBackspaceOrDelete()` in `atoms/src/core/keyboard.ts` (the real backspace/delete handler) fires only `INPUT`, never `TEXTINPUT`, while `updateOnKeyPress()` (character insertion) fires both. Clearing is a deletion, so it should follow the same pattern as backspace/delete, not the insertion path. Will drop the `TEXTINPUT` line from `clear()` and add an event-sequence assertion.

KazuCocoa (approved):

PR #540, opened 2026-08-28
feat: make the Automation session a full WebDriver backend

mykola-mokhnach:
## Summary

Builds out the WebKit `Automation`-domain session (previously used only for JS-dialog handling) into a complete WebDriver-shaped API, so a driver can start a session, get a dedicated browsing context, and drive it end to end without going through the atoms/`Runtime.evaluate` path:

- Window/tab management, navigation, frames, element find/interact, cookies, screenshots, script execution, the W3C Actions API, and JS dialogs, all exposed on `AutomationSession` (`lib/rpc/automation/`, decomposed into a mixin-file module per feature area, mirroring the existing `RemoteDebugger` mixin pattern).
- A dedicated `atoms/automation/*` atom set (`Automation.evaluateJavaScriptFunction`-shaped: raw values in/out, no `{status,value}` envelope) reusing the same `atoms/src/core`/`atoms/src/webdriver` implementations the classic atom set uses, replacing what were originally opaque hand-ported minified JS snippets.
- `lib/rpc/automation/errors.ts`: reclassifies `Automation.*` failures into proper `@appium/base-driver` W3C errors, parsing WebKit's own `Automation.json` `ErrorMessage` names and the W3C error state our atoms embed when they catch a `BotError` (WebKit only relays a thrown error's raw message across the wire).
- Host-side (TypeScript) implicit-wait polling for `findElement`/`findElements` via `asyncbox`'s `waitForCondition`, since the reused locator atoms are one-shot lookups with no built-in retry.
- CI: a "Finalize simulator boot" step (ported from WebDriverAgent) that waits for the simulator's background daemons to settle before running e2e tests, fixing CI timeouts caused by CPU contention right after boot.


lib/mixins/automation.ts
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Starts (or reuses, for the same app) a WebKit `Automation` session and switches the
+ * driven tab to a fresh browsing context it creates - this is the "switch to automation"
+ * moment: from here, W3C commands should go through the returned session instead of atoms.
+ *
+ * @throws {import('@appium/base-driver').errors.SessionNotCreatedError} If no app is
+ *         currently selected, or the selected app isn't Safari with Remote Automation enabled.
+ */
+export async function startAutomationSession(this: RemoteDebugger): Promise<AutomationSession> {
```
KazuCocoa, line 16: So, basically this automation session start when XCUITest driver calls this...?
Nothing will occur this automation stuff without explicit call (== current method keeps working until enabling this)

Just to confirm be expected behavior

lib/rpc/automation/actions.ts
KazuCocoa, line 1: Please preserve `pause.duration` for key, pointer, wheel, and none sources. Today `pointerDown → pause(1000) → pointerUp` collapses to an immediate tap, and a Null source contributes no tick duration. Could we translate pauses and add a non-zero-duration test?

KazuCocoa, line 1: Could input-source state live on `AutomationSession`, with Release Actions wired to `Automation.cancelInteractionSequence`? Recreating these maps per call loses held keys/buttons across Actions commands, and there is currently no release/reset path.

lib/rpc/automation/session.ts
KazuCocoa, line 1: Could this result be recursively converted from `session-node-${sessionId}` handles through `wrapElement()`? As written, `executeScript('return document.body')` exposes WebKit’s internal node shape, including for elements nested in arrays/objects. Please add direct and nested element-result tests.

lib/mixins/automation.ts
KazuCocoa, line 1: If `start()` fails after `ensureStarted()` establishes the protocol session, this object is never retained and cannot be cleaned up. Please catch a start failure, `await session.stop()`, and rethrow.

lib/mixins/automation.ts
mykola-mokhnach: yes. the flow would look like:
- xcuitest switches to safari web view
- xcuitest invokes a mobile method to start an automation session
- xcuitest routes corresponding proxy endpoints to use automation session methods
- after the automation session is finished xcuitest invokes another mobile method to finish the automation session. Switching between contexts does not kill it automatically
- After the automation session is finished endpoints are proxied back to original remote debugger APIs

lib/rpc/automation/elements.ts
KazuCocoa, line 1: Please special-case WebDriver NULL (`\uE000`) here to release all sticky modifiers without emitting text. Because it is absent from `VIRTUAL_KEYS`, the fallback types it literally and modifiers remain active until the final cleanup.

lib/rpc/automation/actions.ts
mykola-mokhnach: Fixed in 3a91f9d: pause.duration is now translated into a WebKit tick duration for key, pointer, wheel, and none sources (a zero-length pause is treated as a no-op and omitted). Added non-zero-duration unit tests for all four source types.

mykola-mokhnach: Fixed in 3a91f9d: held key/button state (pointerInputState/keyInputState) now lives on AutomationSession instead of being recreated per call, so it survives across separate performW3CActions invocations. Added releaseActions(), wired to a new Automation.cancelInteractionSequence primitive (also whitelisted in protocol/index.ts), which cancels the in-progress sequence and clears the held state.

lib/rpc/automation/session.ts
mykola-mokhnach: Fixed in 3a91f9d: added a wrapScriptResult() helper on AutomationSession (the mirror image of the existing toWireArg()) that recursively converts WebKit's session-node-<id> handles back into our W3C element shape, including nested inside arrays/objects. evaluateJavaScriptFunction now applies it to every result, and findElement/findElements/getActiveElement were simplified to rely on it instead of unwrapping manually. Added direct and nested-result tests.

lib/mixins/automation.ts
mykola-mokhnach: Fixed in 3a91f9d: startAutomationSession now catches a start() failure, calls session.stop() on the half-started session (a no-op if the protocol session was never established), and rethrows. Added a test that fails tab creation after a successful handshake and asserts forwardDidClose still gets sent and the session reference stays cleared.

lib/rpc/automation/elements.ts
mykola-mokhnach: Fixed in 3a91f9d: sendKeys now special-cases the WebDriver NULL key ("\uE000") to release every currently-held sticky modifier immediately, without emitting any text, instead of falling through to the literal-character branch. Added a test covering a NULL key mid-string.

lib/rpc/automation/actions.ts
```diff
@@ -0,0 +1,242 @@
...
+  session: AutomationSession,
+  sourceId: string,
+  action: PointerAction | undefined,
+  running: Map<string, PointerRunningState>,
+): StringRecord | null {
+  const state = running.get(sourceId) ?? {};
+  running.set(sourceId, state);
+
+  if (action?.type === 'pointerUp') {
+    running.delete(sourceId);
```
KazuCocoa, line 149: Non-blocking: could `releaseActions()` always call `cancelInteractionSequence()` instead of early-returning based only on the tracked key/pointer maps? WebKit's WebDriver implementation clears its input state and sends `cancelInteractionSequence` for Release Actions. Since wheel/none sources are not represented in these maps, this guard could skip cancellation when host-side held state is empty but WebKit still has sequence state to cancel.

KazuCocoa (approved):
Left one finding to make sure, otherwise lg

PR #543, opened 2026-08-30
fix: Add workarounds for known WebKit automation bugs

mykola-mokhnach:
## Summary

- `callAutomation` now bounds every `Automation.*` command with a configurable `commandTimeoutMs`, converting a wedged connection (WebKit bug [322937](https://bugs.webkit.org/show_bug.cgi?id=322937)) into a clear `TimeoutError` instead of an indefinite hang.
- `AutomationSession.stop()` no longer closes owned browsing contexts by default — that enumerate-and-close step is exactly where the connection tends to wedge on teardown. Now opt-in via `stop({closeAllWindows: true})` / `RemoteDebugger.stopAutomationSession({closeAllWindows: true})`.
- Cookies: read via `Automation.getAllCookies` (not `document.cookie`, which can't see `HttpOnly` cookies or report domain/path/secure/sameSite/expiry). `addCookie` fills in all required fields before calling `Automation.addSingleCookie`, avoiding a rejected call that could also wedge the connection.
- `getWindowRect` falls back to reading the viewport off the DOM when WebKit reports `windowSize` as `{0, 0}`.
- **Native touch tap/click was silently doing nothing.** Root-caused two real bugs by reading WebKit's own source (`Automation.json`, `WebAutomationSession.cpp`, `SimulatedInputDispatcher.cpp`, `WebDriver/Session.cpp`) and cross-checking against a live wire capture:
  1. Every `performInteractionSequence` state we sent omitted `mouseInteraction` (Down/Move/Up). Per WebKit's own docs, "if unmentioned and the interaction cannot be determined through other heuristics, the state is dropped" — and there's no such heuristic in WebKit's C++, only in its reference WebDriver client. We were silently no-oping every tap.
  2. `computeElementLayout` was called with the legacy `LayoutViewport` coordinate system ("should not be used going forward" per WebKit's own docs) instead of `Viewport`, which is what `performInteractionSequence`'s own `origin: 'Viewport'` expects.

  With both fixed, native touch correctly clicks buttons/divs *and* toggles checkbox/radio state — the WebKit-bug-322939 checkbox workaround (routing clicks through a JS atom) was a misdiagnosis of the same root cause and has been removed, along with the now-dead atom it used.
- Investigated `sendKeys` the same way and confirmed the opposite: its request shape already matches WebKit's reference client exactly, and tracing `platformSimulateKeySequence` in `WebAutomationSessionIOS.mm` shows it dispatches raw DOM keyboard events, bypassing the `UITextInput`/IME pipeline iOS Safari actually needs to insert text. That's a structural platform gap, not a fixable request bug — the JS-atom-based `sendKeys()` stays as-is.
- W3C Actions with an `Element` origin resolve the element's on-screen center themselves via `computeElementLayout`, instead of relying on WebKit's own unreliable element-origin resolution (`MoveTargetOutOfBoundsError` even for in-bounds elements).
- `rpc-client.ts`'s `selectPage()` fixed a bug where its fallback timeout could never fire if a page's target was already initialized.
- Generalized `scripts/ci/run-e2e-atoms.mjs` into `scripts/ci/run-e2e.mjs`, taking test file(s) as CLI args; all three `e2e-test*` npm scripts now route through it.
- Added a new per-API isolation e2e suite (`automation-session-api-e2e.spec.ts`, not wired into CI) to make future regressions easier to isolate to a single Automation API.

## Known residual flakiness

`performW3CActions`/`click()` can intermittently throw `TargetOutOfBounds` right after navigation — verified this is a transient WebKit/Simulator timing race (rejected coordinates are well within the viewport, and it reproduces on both our client-resolved and WebKit's own native element-origin resolution). Left unmitigated for now; a retry-on-`TargetOutOfBounds` would be the natural fix if it becomes a real problem.

## Test plan
- [x] Unit tests (`npm test`) — 385 passing
- [x] Full `automation-session-api-e2e.spec.ts` suite run live against a real iOS Simulator (26.5 and 27.0)
- [ ] Atoms e2e (`npm run e2e-test:atoms`)


lib/rpc/automation/actions.ts
```diff
@@ -75,6 +81,38 @@ export async function releaseActions(this: AutomationSession): Promise<void> {
...
+      const {origin} = action;
+      if (origin && origin !== 'viewport' && origin !== 'pointer') {
+        nodeHandles.add(session.unwrapElement(origin as AutomationElement));
+      }
+    }
+  }
+  const centers = new Map<string, {x: number; y: number}>();
+  for (const nodeHandle of nodeHandles) {
+    const layout = await computeLayout.call(session, session.wrapElement(nodeHandle), true, 'LayoutViewport');
+    centers.set(nodeHandle, layout.center);
```
KazuCocoa, line 126: **[P1] Avoid caching element coordinates across scrolls.** Each computeLayout(..., true, ...) call may scroll the page, so resolving a second off-screen element invalidates the cached viewport center of the first before the interaction even starts. Common multi-element actions such as drag-and-drop can therefore click the old location or fail out-of-bounds; earlier wheel/action ticks can also move the DOM. Please preserve element origins or split/recompute coordinates so absolute positions are not cached across viewport changes.

lib/rpc/automation/cookies.ts
```diff
@@ -2,22 +2,71 @@ import type {StringRecord} from '@appium/types';
...
-  await this.callAutomation('addSingleCookie', {browsingContextHandle: this.requireTopLevelHandle(), cookie});
+  const {expiry, ...resolvedCookie} = cookie;
+  if (!resolvedCookie.domain) {
+    resolvedCookie.domain = new URL(await this.getCurrentUrl()).hostname;
+  }
+  resolvedCookie.path ||= '/';
+  resolvedCookie.expires ??= expiry ?? Math.floor(Date.now() / 1000) + DEFAULT_COOKIE_LIFETIME_SECONDS;
+  resolvedCookie.secure ??= false;
+  resolvedCookie.httpOnly ??= false;
+  resolvedCookie.session ??= false;
```
KazuCocoa: **[P2] Preserve session-cookie semantics.** When expiry is omitted, WebDriver requires a session cookie, but this synthesizes a 400-day expiry and defaults session to false. The cookie can persist beyond the browser session and getCookies() will incorrectly return an expiry. WebKit's own WebDriver adapter represents this case as expires: 0 and session: true; please use that mapping when expiry is absent.

lib/rpc/automation/elements.ts
```diff
@@ -38,12 +37,23 @@ export async function findElements(
...
 export async function click(this: AutomationSession, el: AutomationElement): Promise<void> {
-  if ((await this.getTagName(el)) === 'option') {
+  const tagName = await this.getTagName(el);
+  if (tagName === 'option') {
     await selectOptionElement.call(this, el);
     return;
   }
+  if (tagName === 'input' && CHECKABLE_INPUT_TYPES.has((await this.getAttribute(el, 'type'))?.toLowerCase() ?? '')) {
+    await this.evaluateJavaScriptFunction<void>(await getAutomationAtomScript('click'), [el]);
+    return;
```
KazuCocoa: **[P2] Keep the obscured-element check for checkables.** This branch returns before computeLayout checks isObscured. The click atom dispatches synthetic events directly to the requested element rather than hit-testing the element at the click point, so an overlaid checkbox/radio is toggled instead of raising ElementClickInterceptedError. Please run the layout/obscuration check before invoking the atom.

lib/rpc/automation/windows.ts
```diff
@@ -44,15 +44,25 @@ export async function getBrowsingContext(this: AutomationSession): Promise<Autom
...
-  };
+  const x = context.windowOrigin?.x ?? 0;
+  const y = context.windowOrigin?.y ?? 0;
+  const {width, height} = context.windowSize ?? {};
+  if (width && height) {
+    return {x, y, width, height};
+  }
+  const viewport = await this.evaluateJavaScriptFunction<{width: number; height: number}>(
+    'function() { return {width: window.innerWidth, height: window.innerHeight}; }',
+  );
```
KazuCocoa, line 68: **[P2] Read the top-level viewport, not the current frame.** evaluateJavaScriptFunction automatically attaches currentFrameHandle. If getWindowRect() is called while switched into an iframe and WebKit reports a zero window size, this fallback returns the iframe's innerWidth/innerHeight rather than the top-level window rect. Please execute this fallback explicitly in the top-level frame.

package.json
```diff
@@ -50,8 +50,9 @@
     "format:check": "oxfmt -c oxfmt.config.mjs --check .",
     "prepare": "npm run build",
     "test": "node --test --enable-source-maps --test-force-exit --test-timeout=60000 \"./build/test/unit/**/*.spec.js\"",
-    "e2e-test": "node --test --enable-source-maps --test-force-exit --test-concurrency=1 --test-timeout=3600000 \"./build/test/functional/safari-e2e.spec.js\"",
-    "e2e-test:atoms": "node --enable-source-maps scripts/ci/run-e2e-atoms.mjs",
+    "e2e-test": "node --enable-source-maps scripts/ci/run-e2e.mjs \"./build/test/functional/safari-e2e.spec.js\"",
```
KazuCocoa, line 53: **[P2] Do not suppress timeouts in the main E2E suite.** Routing e2e-test through the generalized runner makes every Node testTimeoutFailure exit successfully. Unlike the atoms suite, safari-e2e does not assign intentionally tolerated per-test timeouts; a genuine regression that stalls a functional test can therefore be reported as green when the Node timeout is reached. Please keep timeout suppression scoped to the atoms suite or make it an explicit runner option.

lib/rpc/automation/actions.ts
mykola-mokhnach: Good catch. Fixed in 79c8a81: `resolveElementOriginCenters` now resolves in two passes - scroll every distinct element into view first, then re-read all of them (without scrolling again) once that settles, so every cached center reflects the same final scroll position rather than a mid-scroll snapshot.

Verified live on a real Simulator: a drag between two elements ~2500px apart (further than one viewport) now correctly fails with an honest `MoveTargetOutOfBoundsError` instead of silently using the first element's stale, now-wrong coordinate - which is the right outcome, since both elements genuinely can't be simultaneously reachable from one scroll position given the whole sequence is dispatched to WebKit as a single batched call. A same-viewport two-element drag (no scrolling needed) works reliably. Added a unit test (`should re-read every element center after all scrolling settles...`) covering the staleness scenario directly.

lib/rpc/automation/cookies.ts
mykola-mokhnach: Fixed in 79c8a81: an omitted `expiry` now maps to `expires: 0, session: true`, matching WebKit's own WebDriver adapter, instead of synthesizing a 400-day expiry with `session: false`. Verified live: `getCookies()` on a session cookie now correctly has no `expiry` field at all, while an explicit `expiry` still round-trips as before. Updated the existing unit test to assert the new mapping.

lib/rpc/automation/elements.ts
mykola-mokhnach: This is already resolved as of e874f6b (pushed before this review landed) - the entire checkbox/radio click-atom branch this comment is anchored to has been removed. Turns out the atom routing itself was a misdiagnosis: native touch tapping was silently no-oping for *all* elements due to a missing required `mouseInteraction` field (see the PR description), not something checkbox/radio-specific. Once that was fixed, native touch correctly toggles checkbox/radio state too, so `click()` now always goes through the single native-touch path below, obscured-check included, for every element type.

lib/rpc/automation/windows.ts
mykola-mokhnach: Fixed in 79c8a81: `evaluateJavaScriptFunction` gained a `topLevelOnly` option that skips attaching `currentFrameHandle`, and `getWindowRect`'s DOM-fallback now passes it. Verified live: created a 50px iframe, switched into it, and confirmed `evaluateJavaScriptFunction(..., {topLevelOnly: true})` reads the top-level page's width (~400px) while a call without it reads the iframe's own 50px - so `getWindowRect()` now correctly ignores the current frame.

package.json
mykola-mokhnach: Fixed in 79c8a81: timeout tolerance is now opt-in via a `--tolerate-timeouts` flag on `run-e2e.mjs`, passed only for `e2e-test:atoms`. `e2e-test` (safari-e2e) and `e2e-test:automation` no longer suppress `testTimeoutFailure` - a genuine stall in either will still fail the job.

lib/rpc/automation/actions.ts
```diff
@@ -172,13 +236,13 @@ function buildPointerTickState(
   if (state.location) {
     result.location = state.location;
```
KazuCocoa, line 249: [P1] Do not replay pointer-relative coordinates

When origin is `pointer`, `state.location` stores the relative delta, but this block re-emits that same delta on later pause, pointerDown, or ran-out ticks. WebKit resolves `Pointer` coordinates against its current pointer location on every state, so move(10,20) -> pause -> down advances to (10,20), then (20,40), then presses at (30,60). Please emit location/origin only for the pointerMove tick (a sourceId+duration state can sustain a pause), or normalize tracked locations to absolute viewport coordinates.

lib/rpc/automation/session.ts
```diff
@@ -334,20 +351,29 @@ export class AutomationSession {
...
-      });
+      return await withTimeout(
+        this.rpcClient.send(`Automation.${method}`, {
+          appIdKey: this.appIdKey,
+          pageIdKey: this.automationPageIdKey,
+          senderId: this.sessionId,
+          sessionId: this.sessionId,
+          ...params,
+        }),
+        this.commandTimeoutMs,
```
KazuCocoa, line 368: [P1] Respect the configured page/script timeouts

This 120-second wrapper caps every Automation call, including navigation commands that pass the default `pageLoadTimeoutMs` of 300000 and `executeAsyncScript` when `scriptTimeoutMs` is configured above 120000. A healthy command that WebKit is still allowed to run will therefore be rejected locally two minutes early, making those public timeout settings ineffective. Please let call sites supply the effective timeout, or otherwise ensure the generic command cap cannot undercut a longer protocol-level timeout.

KazuCocoa (approved):
lg entirely though

lib/rpc/automation/session.ts
mykola-mokhnach: This timeout is more like a safety net and should be not reached. Two minutes is a lot in a fast web world. I will think about customizing it if it ever becomes an issue

lib/rpc/automation/actions.ts
mykola-mokhnach: Good catch. Fixed in bc41644: the delta is now resolved to an absolute viewport point immediately (against the last tracked location, defaulting to (0,0) per spec if the source hasn't moved yet), so what gets held/resent on later sustain/pause/pointerDown ticks is a true no-op instead of a re-appliable delta.

Verified live: dispatched `pointerMove(10,20, origin:pointer) -> pause(50) -> pointerDown`, and the resulting `pointerdown` DOM event now fires at exactly (10,20) - previously it would have drifted to (30,60) per your analysis. Added a unit test (`should resolve a pointer-relative origin to an absolute point instead of a resendable delta`) asserting the three emitted ticks all carry the same absolute (10,20)/'Viewport' location.

